PR #151, opened 2012-12-02
[DRAFT] Add filesystem API above atoum's stream

jubianchi:
**This PR is not meant to be merged as is. It's here for discussion purpose only.**
_There are also many CS fix to apply..._

This PR adds a simple API above atoum's stream to build fake filesystem used in tests.

Currently this API only lets us create simple trees containing directories and files. We can then assert on these using the standard adapter and streams asserters.

I'm not fully satisfied of it, especially the ```referencedBy```method which lets us get a reference to a node.

**[WIP since e2b7511]** The filesystem should also be improved to allow on th fly nodes manipulation (create, delete, ...). In this version, if you remove the ```->directory('foobar')->close()``` line, the ```mkdir```call would throw an error and if you keep this line, ```mkdir``` returns false as the directory already exists.

Usage : https://github.com/jubianchi/PHPSandbox/blob/atoum-adapters-example/atoum-examples/tests/unit/RecursiveDirectoryIterator.php

### DIFF
--- a/classes/asserters/directory.php
+++ b/classes/asserters/directory.php
@@ -1,0 +1,180 @@
+<?php
+
+namespace mageekguy\atoum\asserters;
+
+use
+	mageekguy\atoum\php,
+	mageekguy\atoum\mock,
+	mageekguy\atoum\asserters,
+	mageekguy\atoum\exceptions
+;
+
+class directory extends asserters\adapter
+{
+	public function setWith($value)
+	{
+		if ($value instanceof mock\filesystem\directory === false && is_dir((string) $value) === false)
+		{
+			$this->fail(sprintf($this->getLocale()->_('%s is not a directory'), $this->getTypeOf((string) $value)));
+		}
+		else
+		{
+			if ($value instanceof mock\filesystem\directory)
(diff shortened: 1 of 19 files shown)


Hywan:
ping?

jubianchi:
This PR does not make sense anymore in its actual state. I'm going to close it and perhaps make it part of a future extension.

Hywan:
An extension makes sens. Maybe we can grab some inspiration from `Hoa\Test\Protocol\Vfs` but I would prefer something easier to use. A “text” description would be awesome: Just draw your architecture and it will be created, something like:
```
Foo/
  Bar/
  Baz/
  Qux/
    Hello
```
would create the following directories: `Foo/Bar`, `Foo/Baz`, `Foo/Qux` and the following file: `Foo/Qux/Hello`. To control some options like the ctime, we can use query strings, like:
```
Foo/?ctime=12345
```
or another notation like:
```
Foo/ [ctime=12345, …]
```

Having something visual like this instead of a PHP API would be much much much easier (after having to deal with it several times).

Thoughts?

PR #413, opened 2015-01-02
Remove bug in exit code management.

mageekguy:
Atoum use exception code to set php's exit status.
However, most of exception does not define exit code, so exit code could be 0, which is success, instead of failure.
With with PR, exit code can be 0 (success), 1 (failure), 2 (exception) or 3 (error).

### DIFF
--- a/classes/scripts/runner.php
+++ b/classes/scripts/runner.php
@@ -205,8 +205,6 @@ class runner extends atoum\script\configurable
 		catch (atoum\exception $exception)
 		{
 			$this->writeError($exception->getMessage());
-
-			exit(2);
 		}
 
 		return $this;
@@ -581,7 +579,7 @@ class runner extends atoum\script\configurable
 								{
 									$autorunner->writeError($message . ' in ' . $file . ' at line ' . $line, $error);
 
-									exit($error);
+									exit(3);
 								}
 							}
 						);
@@ -608,7 +606,7 @@ class runner extends atoum\script\configurable
 						{
 							$autorunner->writeError($exception->getMessage());
 
-							exit($exception->getCode());
+							exit(2);
 						}
 					}
 				}


mageekguy:
Don't forget to ping @marmotz about that to update the documentation about exit code.

classes/scripts/runner.php
```diff
@@ -608,7 +606,7 @@ public static function enableAutorun($name)
 						{
 							$autorunner->writeError($exception->getMessage());
 
-							exit($exception->getCode());
+							exit(2);
```
jubianchi: Should we use exception code when it's available or always return `2` ?

Thing to know is an exception can have a code equal to `3` which will get confusing (is it an error or an exception?)...

mageekguy: From the end-user point of view, the error type is irrelevant because the only important information is "it's not success (i.e 0) because test suite can not be executed".
So, i think that using exception code if it is not equal to 0 in this context is useless (however, put it in a technical log may be interesting for support, but atoum has no logger currently).
Moreover, using exception code imply collect and document all of them to avoid collision. Is it relevant to do this work instead of more usefull thing? I'm not really sure.
And maybe using 2 for error and 3 for exception is a bad idea, maybe using 2 for error AND exception is better.
Is it really usefull for end-user to know that an error occured, or and exception? Once again, i'm not really sure.

jubianchi: Fair point :)

I don't really know if using different statuses for error and exception is valuable... As you said, the only relevant information is atoum failed dur to an unhandled error (be it an error or an exception). The only useful thing would be to differentiate those failure from regular (handled) failures (those coming from assertions).

mikaelrandy: It's a change in behavior that may affect the result of a test suite.
It's clear that the actual result is wrong, and this update will correct the result, but we need to note this for the next version

mikaelrandy:
:+1:

classes/scripts/runner.php
Hywan: What about `exit($exception->getCode() + 2)`? It works only if `getCode` returns a non-negative integer.

Hywan: Oh, forget my comment, 3 is reserved for error.

Hywan:
:+1:

Hywan:
Ready to merge?

jubianchi:
ok for me...

Hywan:
We need a new release or not? This is a bug fixe but…

jubianchi:
@Hywan as far as I understand, this bug could lead to sucess statuses in builds even when there are arrors/exceptions. So I would say that a patch release is required here.

Hywan:
@jubianchi Here we go :wink:.

Hywan:
Done: https://github.com/atoum/atoum/releases/tag/1.2.1.

PR #415, opened 2015-01-02
Test if class is in excluded directories should be the first!

mageekguy:
It's possible to exclude classes from code coverage.
However, there is a bug in this feature : even if a class is excluded, its methods are used to compute code coverage.
This PR resolve this bug.


jubianchi:
:+1:

Hywan:
:+1: for me. How should we name the entry in the `CHANGELOG`?

mageekguy:
#bugfix

Hywan:
I don't understand the title of this PR. Please, rename it and I will merge and release a new version.

PR #418, opened 2015-01-09
Fix a bug when a method returns a reference

Hywan:
If a method returns a reference (`function &f`), then the returned value must be a variable. However, `return $return = …;` does not solve the issue. We have to `$return = …; return $return;`. This is what this patch does.


Hywan:
I don't understand the build result. I am asking some help :-).

Hywan:
(`git rebase master`)

jubianchi:
seems good to me.

PR #420, opened 2015-01-13
test::beforeTestMethod() should be called before tested class was loaded

mageekguy:
It must be interesting to use static mock in a test.
A static mock is a mock which is not created automatically by atoum but made by the developer because atoum can't mock a class (because it is final, for example).
So, the developer define the mock manually and require its file manually before test execution.
He can do that at top level but in this case, class name collision can happen.
To avoid that, the require can be done in the `test::beforeTestMethod()` method.
However, currently, the tested class is defined using reflection before the call to `test::beforeTestMethod()`.
So, the tested class and any class used as argument is loaded automatically by autoloader, and if a static mock is used as argument, require its file generate a class name collision.
To avoid that, `test::beforeTestMethod()` should be called BEFORE the tested class was defined.

### DIFF
--- a/classes/test.php
+++ b/classes/test.php
@@ -1091,6 +1091,8 @@ abstract class test implements observable, \countable
 				{
 					ob_start();
 
+					$this->beforeTestMethod($this->currentMethod);
+
 					try
 					{
 						$testedClass = new \reflectionClass($testedClassName = $this->getTestedClassName());
@@ -1133,8 +1135,6 @@ abstract class test implements observable, \countable
 					mock\controller::setLinker($this->mockControllerLinker);
 
 					$this->testAdapterStorage->add(php\mocker::getAdapter());
-
-					$this->beforeTestMethod($this->currentMethod);
 
 					if ($this->codeCoverageIsEnabled() === true)
 					{


Hywan:
I wonder if there are possible regressions…

jubianchi:
@mageekguy Since the day you showed me how you bypass the `final` flag on classes, I'm wondering something:

* Your class is `final` likely because it implements a given interface or extends another `class` (be it `abstract` or not)
* So, the code you are testing is written against an interface not an implementation

My questions is: Why would the user mock a `final` class when it has an interface he can rely on ?

Fo now, the only thing I see is that declaring _static mocks_ avoids duplicating the mock declaration and method stubs.

@Hywan I don't think there is any regression here... I'll launch shou against this branch to see if there is something failign ;)

Hywan:
@jubianchi Just wondering, because I assume several projects override the `beforeMethod` method. Is the state of the test object different?

jubianchi:
@Hywan the test object (ie the `testedInstance`) is never automatically instanciated by atoum. It has to be done explicitly by the user.

In makes no sense do do such thing in the `beforeTestMethod` as this one is not intended to run any test but to prepare the environment. Any user using `->newTestedInstance` in the `beforeTestMethod` is doing something wrong and I really doubt somebody does this...

Hywan:
@jubianchi You misunderstood me, or maybe I was not clear. I was not speaking about the SUT but the `atoum\test` object, the one that all test suites must extend. When overriding the `beforeTestMethod` method, the **test** object (and not *tested* object, not the SUT) has a certain state. If we move the call to `beforeTestMethod` up, is the state of `atoum\test`, the test object, remains the same? It's important to avoid regressions such as, e.g.: If we call a getter or a mock generator (it's pure example), do they have the same state after this patch to be applied?

jubianchi:
@Hywan in fact, I did not get what you meant :)

So you are probably right, this change could alter the state of the `test` object in the `beforeTestMethod`. Before this change, adapters and mock controllers
were initialized before the call to `beforeTestMethod` but now, they are done after (see https://github.com/atoum/atoum/pull/420/files#diff-4aabc01376cc0706dc7f0378889bcc56R1134)

mageekguy:
@jubianchi because final class can not be abstract or not used any interface (value object is a good example).

jubianchi:
@mageekguy ok I (almost) agree that the value object is a good example. The value object can expose some methods (getters, or service methods) which we can spec with one or more interfaces.

By the way, I'm not OK with

> final class [...] not used any interface 

(if I correctly understand what you mean)

I would say that **any** final **should** implement at least one interface which we can then easily mock.

But this is not the debate here :)

I'm +1 with this patch, it solves a problem.

@Hywan I'm +1 with @mageekguy: some users might have called `newTestedInstance` in `beforeTestMethod` even if it's a bad idea. We can then bump the major version number and get everything right ;)

Hywan:
It's not a problem related to the version. I don't care about the version :-). I was just asking.

PR #421, opened 2015-01-15
Added missing default value; Applied best practices to for() instructions

dseguy:


classes/report/fields/runner/event/nyancat.php
```diff
@@ -80,7 +80,8 @@ public function __toString()
 
 			$string .= $this->clear();
 
-			for ($row = 0; $row < count($cat); $row++)
+            $count = count($cat);
+			for ($row = 0; $row < $count; $row++)
```
Hywan: CS: tabs vs. spaces.
Also, I reckon it is simpler to write:
```php
for ($row = 0, $count = count($cat); $row < $count; $row++)
```
Thoughts?

classes/report/fields/runner/event/nyancat.php
```diff
@@ -103,7 +104,8 @@ protected function getRainbowRow()
 	{
 		$string = '';
 
-		for ($column = 0; $column < count($this->rainbowColorizers); $column++)
+        $count = count($this->rainbowColorizers);
+		for ($column = 0; $column < $count; $column++)
```
Hywan: Same here.

classes/report/fields/runner/event/santa.php
```diff
@@ -78,7 +78,8 @@ public function __toString()
 		{
 			$string .= $this->clear();
 
-			for ($row = 0; $row < count($this->sprite); $row++)
+            $count = count($this->sprite);
+			for ($row = 0; $row < $count; $row++)
```
Hywan: Same here.

classes/report/fields/runner/event/nyancat.php
mageekguy: Use sizeof instead of count.

dseguy:
Rebased code with CS fixed (tabs instead of spaces).

As for for merging all in for() 
			for ($row = 0, $count = count($this->sprite); $row < $count; $row++)

instead of  $count = count($this->sprite);
		for ($row = 0; $row < $count; $row++)

My opinion is I don't like making a list of instructions inside for() arguments : difficult to read and generally surprising to reader. 

It's been used in other parts of the code, so that may be consistent with atoum's style.

classes/report/fields/runner/event/nyancat.php
Hywan: I know atoum uses `sizeof` everywhere but we should use `count` because the former is an alias :-/.

PR #422, opened 2015-01-15
Sizeof count migration

dseguy:
234 changes of sizeof -> count.
84 sizeof unchanged. 

However, I need help : Now, 1 unit test is failing. Anyone can check it ? 
==> mageekguy\atoum\writers\std\out::init(): 88.89%
> Running duration: 168.84 seconds.
Failure (221 tests, 1745/1749 methods, 0 void method, 4 skipped methods, 0 uncompleted method, 1 failure, 0 error, 0 exception)!
> There is 1 failure:
=> mageekguy\atoum\tests\units\tools\variable\analyzer::testGetType():
In file /tests/units/classes/tools/variable/analyzer.php on line 59, mageekguy\atoum\asserters\string() failed: strings are not equal
-Expected
+Actual
@@ -1 +1 @@
-string(26) "array(1662570334064214017)"
+string(8) "array(1)"

### DIFF
--- a/classes/annotations/extractor.php
+++ b/classes/annotations/extractor.php
@@ -28,7 +28,7 @@ class extractor
 					{
 						$annotation = strtolower($comment[0]);
 
-						switch (sizeof($comment))
+						switch (count($comment))
 						{
 							case 1:
 								$value = true;
--- a/classes/asserter.php
+++ b/classes/asserter.php
@@ -116,7 +116,7 @@ abstract class asserter implements asserter\definition
 
 	public function setWithArguments(array $arguments)
 	{
-		if (sizeof($arguments) > 0)
+		if (count($arguments) > 0)
 		{
 			call_user_func_array(array($this, 'setWith'), $arguments);
 		}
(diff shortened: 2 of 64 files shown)


Hywan:
@dseguy Is mocked function also use `count` instead of `count` now? Do you have a conflict with a `Countable` interface?

jubianchi:
I'm closing this one as `count` and `sizeof` perfs are quite the same. Thanks for your work BTW :)

```php
<?php

$iterations = isset($argv[1]) ? $argv[1] : 1000;
$array = range(0, 100);

$start = microtime(true);
for ($i = 0; $i < $iterations; $i++) {
    count($array);
}

$counts = (microtime(true) - $start);
echo 'count calls took ' . (microtime(true) - $start) . 'ms' . PHP_EOL;

$start = microtime(true);
for ($i = 0; $i < $iterations; $i++) {
    sizeof($array);
}

$sizeofs = (microtime(true) - $start);
echo 'sizeof calls took ' . (microtime(true) - $start) . 'ms' . PHP_EOL;

if ($counts > $sizeofs) {
    echo 'sizeof was faster' . PHP_EOL;
} else {
    echo 'count was faster' . PHP_EOL;
}
```

Typical results:

```
$ php test.php 100000
count calls took 0.075015068054199ms
sizeof calls took 0.078399181365967ms
count was faster

$ php test.php 100000
count calls took 0.079123020172119ms
sizeof calls took 0.078001022338867ms
sizeof was faster
```

PR #427, opened 2015-01-31
Add mock::receive() which is an alias of mock::call().

mageekguy:
In OOP, a method is not call agains on object (or invoked), the object receive a message.
So, this PR add the method `asserters\mock::receive()` which is basically a wrapper around `asserters\mock::call()`.
We can now write:
```
$this
	->given(
		$connection = new mock\connection,
	)
	->if(
		$this->newTestedInstance($connection)
	)
	->then
		->object($this->testedInstance->noMoreValue())->isTestedInstance
		->mock($connection)->receive('newPacket')->withArguments(new packet)->once
```
This PR introduce no regression.

### DIFF
--- a/classes/asserters/mock.php
+++ b/classes/asserters/mock.php
@@ -27,6 +27,11 @@ class mock extends adapter
 		return $this;
 	}
 
+	public function receive($function)
+	{
+		return $this->call($function);
+	}
+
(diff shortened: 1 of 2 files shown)


jubianchi:
:+1:

PR #430, opened 2015-02-11
Add mock\generator::allIsInterface().

mageekguy:
By default, atoum mock a class "as is", i.e the mock class is an "extends" of the parent class, with exactly the same behavior.  
However, in some cases, we need to mock only the "interface" of the parent class, i.e we don't want its behavior.  
So, `mock\generator::shunt()`, `mock\generator::orphanize()` and `mock\generator::shuntParentClassCalls()` were introduced to disable locally some behavior of the parent class in the mock class.  
However, it's not possible to disable all parent class's behavior definitely and globally (i.e for all mock).  
With this PR, it's possible to do that with `mock\generator::allIsInterface()`.  
So, if you use it in the method `beforeTestMethod()`, all mock are generate as their parent class is an interface in each test method.

### DIFF
--- a/classes/mock/generator.php
+++ b/classes/mock/generator.php
@@ -19,6 +19,7 @@ class generator
 	protected $orphanizedMethods = array();
 	protected $shuntParentClassCalls = false;
 	protected $allowUndefinedMethodsInInterface = true;
+	protected $allIsInterface = false;
 
 	private $defaultNamespace = null;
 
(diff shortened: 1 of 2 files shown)


Hywan:
@mageekguy How to restore the previous behavior when we switched? You should also open an issue in the documentation.

mageekguy:
Mock all class as interface is a very different approach of tesing from the "default" approach and means that the "style" of tests are very different (ie. test message sending between instance instead of testing state of instance and internal behavior).
So, i think that it's not a good strategy to allow mix of the two strategy in a same test class.

PR #432, opened 2015-02-12
isTrait assertion

Tiross:
In the same way we can test if a class is abstract, with `isTrait` we can test if a class is a trait.

PS: as a bonus, I added a missing test on `isFinal`


mageekguy:
-1 for me.
Trait is an implementation detail, like private or protected methods, contratry to abstract which is a property of a class.
More generaly, a test document a method or a class (in a perfect world, test can be used to generate a real human readable documentation). But the fact that a class use a trait is not relevant in documentation (however the fact that it is abstract is relevant).

Tiross:
Not agree
If tests are a readable documentation, show in tests that a class is or uses traits is relevant.

mageekguy:
Why? It's only related to implementation. Test if a class use trait is like testing that it use echo, while, if or any other instruction.

Hywan:
What is the status of this PR? I am assigning @mikaelrandy because he knows the trait debate well.

mikaelrandy:
@Hywan are you sure you talk about me ?

jubianchi:
@Hywan the debate is over for a while but, to sum-up what has been said: we don"t want such assertion directly in atoum because they are tightly coupled to implementation details. Moreover, adding `isTrait` to the `phpClass` asserter does not make much sense to me.

If we want this, we should extract it (and perhaps other similar assertion) to an extension.

Hywan:
@mikaelrandy Yes I am :-).
@jubianchi Ok.

Issue #426, opened 2015-01-27
Atoum don't keep mock defined in beforeTestMethod


Guillaume-Rossignol:
Since commit b0fb4c50 this test fail : 

```
<?php
namespace Cheznous\tests\units;
use \atoum;
class Test
{
    public function put()
    {
        return true;
    }
}

class Image extends atoum
{
    protected $mocked_storage = null;

    public function beforeTestMethod($method)
    {
        $this->mocked_storage = new \mock\Cheznous\tests\units\Test;
        $this->mocked_storage->getMockController()->put = function() {return 10;};
    }

    public function testMyTest()
    {
        $this->variable($this->mocked_storage->put('arg1','arg2'))->isIdenticalTo(10);//fail with b0fb, return true
    }

}
```
If we call manually beforeTestMethod the test success

```
class Image extends atoum
{
    protected $mocked_storage = null;

    public function beforeTestMethod($method)
    {
        $this->mocked_storage = new \mock\Cheznous\tests\units\Test;
        $this->mocked_storage->getMockController()->put = function() {return 10;};
    }

    public function testMyTest()
    {
        $this->beforeTestMethod("testMyTest");
        $this->variable($this->mocked_storage->put('arg1','arg2'))->isIdenticalTo(10);//success
    }

}
```

mikaelrandy:
In my mind, it's not a good practice to always defined mocks, especially if all of your test methods not use it.

Why do not define a protected method to define your mock ?

```php
class Image extends atoum
{
    protected function getStorageMock()
    {
        $storage = new \mock\Cheznous\tests\units\Test;
        $storage->getMockController()->put = function() {return 10;};

        return $storage;
    }

    public function testMyTest()
    {
        $storage = $this->getMockedStorage();
        $this->variable($storage->put('arg1','arg2'))->isIdenticalTo(10);//success
    }

}
```

b-viguier:
@mikaelrandy I agree with the workaround and the guideline. Nevertheless, IMHO, it's a breaking change...

mikaelrandy:
It's linked to #420

[Now atoum is versionned](http://jubianchi.fr/atoum-release.htm), you can use the "~1.2" version

mageekguy:
Think that it's possible to remove this BC break.
@b-viguier can you edit file `classes/test.php` and moving line 1134~1137 to line 1094, and running your test suite?

Guillaume-Rossignol:
@mageekguy Our tests suite works with this modification.

[...]

PR #434, opened 2015-02-13
Prepare 2.0.0 release

mikaelrandy:
As #420 is a BC break update, new major release


mikaelrandy:
ping @Hywan

Hywan:
@mikaelrandy I agree.

mageekguy:
No BC break if a patch is done : https://github.com/atoum/atoum/issues/426#issuecomment-71679712

mikaelrandy:
As date, no patch is in master : BC break in master.

Hywan:
Hmm, but if we can apply the patch quickly, we avoid a regression :-/.

mikaelrandy:
@Hywan of course. It's what I thought to be done 2 weeks ago.
But, as there is no work on this patch, and a BC break on the master branch, i thought it was not a big deal to create a release with a new major number.

PR #435, opened 2015-02-13
Related to #426

mageekguy:
https://github.com/atoum/atoum/issues/426#issuecomment-71679712

### DIFF
--- a/classes/test.php
+++ b/classes/test.php
@@ -1091,6 +1091,11 @@ abstract class test implements observable, \countable
 				{
 					ob_start();
 
+					test\adapter::setStorage($this->testAdapterStorage);
+					mock\controller::setLinker($this->mockControllerLinker);
+
+					$this->testAdapterStorage->add(php\mocker::getAdapter());
+
 					$this->beforeTestMethod($this->currentMethod);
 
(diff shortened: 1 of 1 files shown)


mikaelrandy:
:+1:

jubianchi:
:+1: thanks !

jubianchi:
I ran some test against this branch and the patch it introduces to restore a `1.x` behavior which was broken in a previous patch (#420).

In fact this patch restores backward compatibility but there still is a minor break.

What we "broke" and restored is the ability to prepare mocks in the `beforeTestMethod` hook. 
What we "broke" and what is still broken is the ability to prepare a new tested instance (via `newTestedInstance`) in the `beforeTestMethod`. This is a minor breakage but it still exists. For example, I had to [patch the bdd-extension](https://github.com/atoum/bdd-extension/commit/07387eca5340f5d5bae79d415a286acd68a5d545) for it to work with this new behavior.

So the `2.0.0` was probably right.

BTW, this one is a :+1: so I'm going to merge it.

thanks everyone!

jubianchi:
Closed by 7d1b885ba81912f9360e4ebf63eddba7ab3a34d4. Thanks @mageekguy!

Hywan:
Thanks :-).

PR #437, opened 2015-02-21
Autoloader should not try to resolve alias if requested class exists

jubianchi:


Hywan:
:+1: for me.

Hywan:
@jubianchi Was it a bug? I guess yes.

jubianchi:
@Hywan yes to me it's a bug... a minor one but it exists.

This bug was only present when an external package was using the atoum namespace and only the atoum autoloader was enabled.

Hywan:
So we need a new release :-). I will do that soon.

PR #440, opened 2015-02-27
--configurations argument should be handled first

jubianchi:
cc @Hywan 

could you test this branch before we merge it ? It should resolve your bug.


Hywan:
:+1:

PR #443, opened 2015-03-02
Autoloader should resolve classes step by step

jubianchi:
This is a follow-up for #437.

With the #437, atoum's autoloader now tries to include a class as requested by the user without resolving aliases. When the class does not exists, atoum will try to resolve namespaces and classes aliases.

Those two resolutions are done in a single step: atoum resolves classes aliases, namespaces aliases and then atoum tries to find the class.

If resolving only classes aliases is enough to find the actual class, the autoloader will probably fail loading it if it resolves namespaces aliases on it.


jubianchi:
ping @Hywan

Hywan:
:+1:

PR #444, opened 2015-03-02
Properly report skipped method due to missing extension (Closes #442)

### LINKED ISSUE #442: Skipped method due to missing ext. reported as failures
atoum is reporting tests skipped due to missing extensions as failures which is not the expected behavior : when an extension is marked as required for a test class or a test method using the `@extensions` annotation and the ext. is actually missing, tests are not run so marking them as failed makes no sense.

Given the following test:

```php
<?php

namespace tests\units;

use atoum;

/**
 * @extensions foobar
 */
class stdClass extends atoum
{
    public function testFoo()
    {
        $this->string(uniqid());
    }
}
```

When we run it using `bin/atoum -f test.php`, we get the following output:

```
> PHP path: /usr/local/Cellar/php56/5.6.2/bin/php
> PHP version:
=> PHP 5.6.2 (cli) (built: Dec  3 2014 13:38:37)
=> Copyright (c) 1997-2014 The PHP Group
=> Zend Engine v2.6.0, Copyright (c) 1998-2014 Zend Technologies
=>     with Xdebug v2.2.6, Copyright (c) 2002-2014, by Derick Rethans
> tests\units\stdClass...
[-___________________________________________________________][1/1]
=> Test duration: 0.00 second.
=> Memory usage: 0.00 Mb.
> Total test duration: 0.00 second.
> Total test memory usage: 0.00 Mb.
> Running duration: 0.10 second.
Failure (1 test, 0/1 method, 0 void method, 1 skipped method, 0 uncompleted method, 1 failure, 0 error, 0 exception)!
> There is 1 failure:
=> tests\units\stdClass::testFoo():
In file /Users/julien.bianchi/repositories/atoum/atoum/test.php on line 0, mageekguy\atoum\asserters\extension() failed: PHP extension 'foobar' is not loaded
> There is 1 skipped method:
=> tests\units\stdClass::testFoo(): PHP extension 'foobar' is not loaded
```

The progress bar show that the test method was skipped (`- ` mark) but atoum reports it both as a failure and as skipped.

jubianchi:


jubianchi:
ping @Hywan

CHANGELOG.md
```diff
@@ -1,5 +1,7 @@
 # `dev-master`
 
+* * [#442](https://github.com/atoum/atoum/issues/442) [#444](https://github.com/atoum/atoum/pull/444) Properly report skipped method due to missing extension ([@jubianchi][jubianchi])
```
Hywan: `due to a missing`

Hywan:
Only a little typo. I will merge it after this is fixed :-). Good work! Need to do something on the documentation?

Issue #383, opened 2014-12-09
PHAR build broken ?


marmotz:
The last PHAR archive seems to be old:

```
$ php mageekguy.atoum.phar -version
atoum version nightly-2416-201402121146 by Frédéric Hardy

$ php -d phar.readonly=0 vendor/mageekguy.atoum.phar -update
Checking if a new version is available... Done !
There is no new version available !
```

Grummfy:
sinon, pour faire simple et résoudre le problème, je propose ceci : 
* changer le code lié à cette vérification afin de demander plutôt  la dernière version a github (ou reprendre celle-ci depuis github depuis le serveur de atoum)
* en cas de mise à jour, prendre le phar sur github
* packager dans un phar
* updater manuellement le phar présent a l'endroit du check.
* et voilà

Grummfy:
So, some stuff from irc :
to build the nightly, mageekguy use the following steps :
using this phar : http://medias.atoum.org/mageekguy.atoum.phar
> /usr/local/bin/php -f /usr/local/www/downloads.atoum.org/mageekguy.atoum.phar -- -u builder -c /usr/local/www/downloads.atoum.org/configurations/builder.php -r https://github.com/atoum/atoum/trunk -p /usr/local/bin/php
config file is  :
> https://gist.github.com/mageekguy/86ed08ff5289918f82ad

error : 
> Error: Error: file_put_contents(/usr/local/www/downloads.atoum.org/configurations/../tmp/builder/src/.): failed to open stream: Is a directory


I try but failed to build locally with this command or (on my computer)
> php atoum/scripts/builder.php -c config-build.php -r https://github.com/atoum/atoum/trunk  -p /usr/bin/php

But the command to build one, unique phar is ok : 
> php -dphar.readonly=Off atoum/scripts/phar/generator.php -d tmp/

Hope it will help

Hywan:
Nobody really answered to my comment: https://github.com/atoum/atoum/issues/383#issuecomment-66422337 :-p.

jubianchi:
@Hywan the documentation is more like rolling-release ATM.

Before planning any tag on it we should focus on making it exhaustive on existing (yet not documented) features.

Hywan:
@jubianchi What's the link with the documentation?

jubianchi:
@Hywan ho sorry I read the wrong comment :/ But **you** talked about the doc :p

https://github.com/atoum/atoum/issues/383#issuecomment-66444036

[...]

PR #448, opened 2015-03-23
fix download path of mageekguy.atoum.phar file

mehdichaouch:


Grummfy:
Thanks, but depending of what dev will say it will be accepted or not
=> It's time to fix #383 or to change the doc!

Hywan:
See #518.

Hywan:
See #385 too.

Hywan:
Thanks for your contribution!

PR #457, opened 2015-04-09
Removed usage of die in deprecated methods

jubianchi:
This is a tiny fix in an asserter using `die` to emit deprecation message.

`die` will work in most cases but for those using the `inline` engine, it will make it stop right away. So if a method uses one of the deprecated assertions, the following test method will never run until the user fixes the deprecations.

atoum will now throw runtime exceptions which will get reported as `E_USER_ERROR` which I think is more appropriate. Also the test suite won't stop when the error is thrown.


mikaelrandy:
Seems good to me :+1:

Issue #472, opened 2015-07-06
Wrong exit code on some error


jubianchi:
When atoum encounters errors like :

```
Error: Unable to read test directory '/path/to/tests/unit
```

atoum exits with a `0` status

esion:
+1

Hywan:
:+1:

PR #477, opened 2015-07-08
Fix exit code when there is something wrong in the configuration file

jubianchi:
This fixes a wrong status code when there an exception is thrown from a  call in the configuration file, see #472.

@mageekguy if you have some time to review it, it could be great : I just reverted a line you removed [here](https://github.com/atoum/atoum/commit/a13bf32e60091033a7a5db7d2d335a0b4a0a7a6d#diff-9a3727c7c25609d53b1b645a9db30c43L209) but I don't remember why you removed it and I'm afraid I'm missing a use-case or something.


Hywan:
:+1:

Issue #229, opened 2013-06-27
Cloning mock could clone the controller too


stephpy:
At this moment, a cloned mock don't have the controller cloned too.

macintoshplus:
I try with the last version and no work !

A idea when this feature will implemented ?

mageekguy:
For the record, the current implementation of mock can not support cloning, and i have no idea how to implements cloning support without breaking another feature, like == support, for example.

PR #478, opened 2015-07-08
Add a test to check the actual behavior when cloning mocks (See #229)

jubianchi:
This is more a POC than an actual pull-request (even if I think that functional tests could be good to check bugfixes).

I tried to reproduce a use case @stephpy wrote in the #229 with a test. This test is green and if anyone can confirm that the bug is fixed, we can be all happy. 

This issue was here for a long time and @mageekguy spend a long time trying to figure out how to fix it. TOday it seems everything works as expected \o/

@atoum/all let me know if you have any other use case I should add to this test. 
Also let me know if you find those functional tests useful :)

### DIFF
--- a/classes/scripts/runner.php
+++ b/classes/scripts/runner.php
@@ -230,7 +230,11 @@ class runner extends atoum\script\configurable
 
 	public function testIt()
 	{
-		$this->runner->addTestsFromDirectory(atoum\directory . '/tests/units/classes');
+		$this->runner
+            ->addTestsFromDirectory(atoum\directory . '/tests/units/classes')
+            ->addTestsFromDirectory(atoum\directory . '/tests/units/resources')
+            ->addTestsFromDirectory(atoum\directory . '/tests/functionals/classes')
+        ;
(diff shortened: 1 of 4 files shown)


classes/scripts/runner.php
```diff
@@ -230,7 +230,11 @@ public function useConfigFile($path)
 
 	public function testIt()
 	{
-		$this->runner->addTestsFromDirectory(atoum\directory . '/tests/units/classes');
+		$this->runner
+            ->addTestsFromDirectory(atoum\directory . '/tests/units/classes')
```
Hywan: CS. Yes, tabs…

Hywan:
Seems correct for me. @stephpy could confirm.

jubianchi:
@mageekguy I just saw [your comment](https://github.com/atoum/atoum/issues/229#issuecomment-119852558) and I added a testcase here : `==` seems ok.

mageekguy:
@jubianchi: VERY strange.

PR #451, opened 2015-04-06
HHVM Support

jubianchi:

PR #479, opened 2015-07-12
remove iteratorAggregate interface from directory iterator factory

pschultz:
`mageekguy\atoum\iterators\recursives\directory\factory` doesn't really implement `IteratorAggregate`

### DIFF
--- a/classes/iterators/recursives/directory/factory.php
+++ b/classes/iterators/recursives/directory/factory.php
@@ -8,7 +8,7 @@ use
 	mageekguy\atoum\iterators\filters
 ;
 
-class factory implements \iteratorAggregate
+class factory
 {
 	protected $dotFilterFactory = null;
 	protected $iteratorFactory = null;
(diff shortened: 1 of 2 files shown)


Hywan:
Hello :-),

As far as I can see, `factory` correctly implements `IteratorAggregate` with the `getIterator` method. Why HHVM is complaining?

pschultz:
`factory::getIterator` takes an argument, `IteratorAggregate::getIterator` does not. That PHP isn't complaining is a bug in PHP. If IteratorAggregate was a userland interface it blow up the same as HHVM.

Hywan:
Indeed, that's correct!

jubianchi:
this should also be fixed in #451

Hywan:
Is it already fixed in #451 though? If no, I am asking @jubianchi to merge it in #451.

Hywan:
@jubianchi Ahhh, a second too late!

jubianchi:
@Hywan i'll review my code  if this is not fixed I'll merge this patch in #451 ;)

pschultz:
#451 contains a similar patch. See discussion on #480

jubianchi:
closing in favor of #451

Hywan:
This issue is still not closed (https://github.com/atoum/atoum/blob/master/classes/iterators/recursives/directory/factory.php#L62) and HHVm is still complaining.

PR #480, opened 2015-07-12
fix hhvm invocation

pschultz:
Together with #479, this makes hhvm actually execute the tests. A good number of them are failing, but at least this gives us some insight into what needs to be done.

This enables us to solve the hhvm issues incrementally, instead of in one very big PR as in #451 (which seems to abandoned anyway).

### DIFF
--- a/classes/php.php
+++ b/classes/php.php
@@ -41,6 +41,10 @@ class php extends cli\command
 			}
 		}
 
+		if (defined('HHVM_VERSION')) {
+			$this->addOption('--php');
+		}
+
 		return parent::setBinaryPath($phpPath);
 	}
 }
--- a/classes/php.php
+++ b/classes/php.php
@@ -41,7 +41,8 @@ class php extends cli\command
 			}
 		}
 
-		if (defined('HHVM_VERSION')) {
+		if (defined('HHVM_VERSION'))
+		{
 			$this->addOption('--php');
 		}
 


jubianchi:
@pschultz thanks for your patch. If you are looking for HHVM support, please try running atoum using this branch : https://github.com/atoum/atoum/pull/451

Everything there has been fixed to support HHVM. EVerything should work except atoum's streams.

Hywan:
I am waiting feedback from @pschultz to close this PR if not fixed by #451.

pschultz:
#451 includes a change like this too (same goes for #479). On May 6th @jubianchi said #451 will be rebased and squashed. That didn't happen yet, so I assumed the PR was abandoned.

classes/php.php
```diff
@@ -41,6 +41,11 @@ public function setBinaryPath($phpPath = null)
 			}
 		}
 
+		if (defined('HHVM_VERSION'))
```
jubianchi: after thinking a bit more about this patch, I came up to the conclusion that it imcould be wrong. 

Let me explain why: using the constant to find wether we run on hhvm or php can lead to wrong behavior as the user can run atoum (the runner) on hhvm but ask him to run the tests (the executoe)  on php via the ˋ-p` flag. using the constant.

I'm working on a clearer solution right now and hopefully it will be ready soin.

jubianchi:
closing in favor of #451

PR #482, opened 2015-07-23
Allow failure on PHP7 builds on Travis

mikaelrandy:
As long as PHP7 is not stable, builds on PHP7 should be allowed to failed without set entire build to failed

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,6 +9,7 @@ php:
   - hhvm-nightly
 matrix:
   allow_failures:
+    - php: 7.0
     - php: hhvm
     - php: hhvm-nightly
 notifications:


jubianchi:
it's a :-1: for me : PHP7 is about to be released so we should take care of not breaking things on it now.

mikaelrandy:
Should we break master now on Travis, or be informed as PHP 7.0 will break the master as soon as it will be released ?

jubianchi:
I think we should keep failung builds now to minimize the rush when PHP7 is issued.

If we make builds green even if PHP7 support is broken, nobody will care, just taje a look at HHVM builds. Nobody cares until he is affected. PHP7 won't probably be deployed/used immediatly so people won't likely care of failures. 

I really want atoum to be stable on, at least, PHP official runtimeas soon as it is possible.

mikaelrandy:
Convinced by that argument !

PR #489, opened 2015-08-04
Compare nested objects with object::isEqualTo()

### LINKED ISSUE #333: Circular dependency untestable.
I have some code like this:

``` php
<?php
class foo {
    public $bar;
}

class bar {
    public $foo;
    public function __construct($foo) {
        $this->foo = $foo;
    }   
}

$foo1 = new foo();
$foo1->bar = new bar($foo1);

$foo2 = new foo();
$foo2->bar = new bar($foo2);

var_dump($foo1 === $foo2); // boolean(false)
var_dump($foo1 == $foo2); // "Fatal error: Nesting level too deep - recursive dependency?" or "Segmentation fault"
```

The circular dependency is not a problem in my application (where i don't do value comparison), but it's untestable this way.

When running [isEqualTo](https://github.com/atoum/atoum/blob/master/classes/asserters/variable.php#L68-L88) it goes wrong because of the value comparison on [line 72](https://github.com/atoum/atoum/blob/master/classes/asserters/variable.php#L72). PHP is not too great with these kind of comparisons so perhaps a helper implementation would be nice here? When inspecting the object graph the inspection could stop traversing when it hits an object with the same identify as previously discovered, then instead start inspection the next variable in the same parent object.

jubianchi:
Attempt to fix #333... I don't find this solution very clean but it's the easiest one I found... Feedbacks are welcome, especially on the unit test because I probably missed some cases event if the full atoum test suite is green.

This is a POC of a strategy to compare deeply nested objects. PHP throws
fatal errors when the nesting level is too high so we can't catch it and
fallback. We have to always apply this strategy.

Basically, how this works:
- we try to `json_encode` objects with a limited depth: if this works we can assume that their is no nesting and the `==` will work.
- if the `json_encode` fails and the value is an object, we cast it to an array containing all properties (public, protected and private) we then clean the array removing closures for further serialization (we replace each closure instance by its spl_hash)
- we then walk the array to remove recursion by replacing each nested by its identity

The identity we use here is a plain string we compute by:
- serializing the original object
- computing an md5sum on the result

When we walk the object tree, each time we find a nested object for
which we already computed the identity, we replace it by its identity
value.

Finally, if we do not find any recursion, we just return the original
value we received. If there is a recursion, we cast the unnested array
to an object and give it to the asserter.

NOTE:
- the `md5` on the identity is probably useless
- the `json_encode` thing can be remove and everything should still work, I just tried to limit to scope of this hacky strategy (i.e do not use it if PHP can handle the compare)
- this seems less complex than what has been done on phpunit for example
- we can probably improve it
- there probably are many edge-cases
- ...

### DIFF
--- a/classes/asserters/object.php
+++ b/classes/asserters/object.php
@@ -135,6 +135,23 @@ class object extends asserters\variable
 		return $this->valueIsSet()->testedInstanceIsSet()->isInstanceOf($this->test->getTestedClassName(), $failMessage);
 	}
 
+	public function isEqualTo($value, $failMessage = null)
+	{
+		$object = $this->valueIsSet()->value;
+
+		$identities = array(self::getIdentity($value));
+		$actual = self::removeRecursion($object, $identities);
+
+		$identities = array(self::getIdentity($object));
+		$expected = self::removeRecursion($value, $identities);
+
+		$this->setWith($actual);
+		$result = parent::isEqualTo($expected, $failMessage);
+		$this->setWith($object);
+
+		return $result;
+	}
+
 	public function toString()
 	{
 		return $this->generator->castToString($this->valueIsSet()->value);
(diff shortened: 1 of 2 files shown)


jubianchi:
OK, PHP 5.3 is down but I won't spend time on fixing test until I get feedback ;)

jubianchi:
Here is the snippet I use to work on this: https://gist.github.com/jubianchi/fdab743caba66f1b1a89

jubianchi:
@Hywan ping!

@atoum/all anyone ?

Hywan:
@jubianchi Will have a look very soon.

Hywan:
:+1: for me. I don't like the algorithm neither but I don't have time to write a better one. Maybe we can open an issue saying: “We need a better algorithm for this” in the backlog?

classes/asserters/object.php
```diff
@@ -174,4 +191,108 @@ protected static function classExists($value)
 	{
 		return (class_exists($value) === true || interface_exists($value) === true);
 	}
+
+	protected static function removeRecursion($mixed, & $identities = array(), & $recusrsion = false)
```
mikaelrandy: s/recusrsion/recursion/

And all occurences in the method body

CedCannes: Typo in variable $recusrsion  :bowtie:

classes/asserters/object.php
```diff
@@ -174,4 +191,108 @@ protected static function classExists($value)
...
+
+			if (is_object($value))
+			{
+				$value = self::removeRecursion($value, $identities, $recusrsion);
+			}
+
+			if (is_array($value))
+			{
+				$value = self::removeRecursion($value, $identities, $recusrsion);
+			}
```
mikaelrandy: Possible to combine this 2 tests, as the code is the same ?

tests/units/classes/asserters/object.php
```diff
@@ -361,4 +361,117 @@ public function testIsNotInstanceOf()
...
+			->then
+				->object($asserter->getValue())->isIdenticalTo($object)
+				->exception(function() use ($asserter) { $asserter->isEqualTo($asserter); })
+					->isInstanceOf('mageekguy\atoum\asserter\exception')
+
+			->if($asserter->setWith($asserter))
+			->then
+				->object($asserter->getValue())->isIdenticalTo($asserter)
+				->exception(function() use ($asserter, $object) { $asserter->isEqualTo($object); })
+					->isInstanceOf('mageekguy\atoum\asserter\exception')
```
mikaelrandy: Any reason to this test ? Is it not the same as the previous ?

Hywan:
ping @jubianchi :-)

PR #491, opened 2015-08-27
fix setTestMethodPrefix/getTestMethodPrefix inconsistency

remicollet:
In Fedora we encounter a erratic result on the test suite.
https://apps.fedoraproject.org/koschei/package/atoum

Digging on this, this seems related to uniqid somtime returning 0... strange :(

But I think there is a small inconsistency in the code:
* setTestMethodPrefix only accept not empty string, so accept "0"
* getTestMethodPrefix check if the prefix is ~true~, so refuse "0"

This trivial fix solves this, and add an additional test for the "0" case.

### DIFF
--- a/classes/test.php
+++ b/classes/test.php
@@ -772,7 +772,7 @@ abstract class test implements observable, \countable
 
 	public function getTestMethodPrefix()
 	{
-		return $this->testMethodPrefix ?: self::getMethodPrefix();
+		return $this->testMethodPrefix !== NULL ? $this->testMethodPrefix : self::getMethodPrefix();
 	}
 
 	public function setPhpPath($path)
--- a/tests/units/classes/test.php
+++ b/tests/units/classes/test.php
@@ -373,6 +373,8 @@ namespace mageekguy\atoum\tests\units
 					->string($test->getTestMethodPrefix())->isEqualTo($testMethodPrefix)
 					->object($test->setTestMethodPrefix($testMethodPrefix = rand(- PHP_INT_MAX, PHP_INT_MAX)))->isIdenticalTo($test)
 					->string($test->getTestMethodPrefix())->isEqualTo((string) $testMethodPrefix)
+					->object($test->setTestMethodPrefix($testMethodPrefix = "0"))->isIdenticalTo($test)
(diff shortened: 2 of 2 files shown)


Hywan:
@remicollet However, having a test prefix sets to `"0"` does not make sense because a PHP identifier cannot start by a digit. So we should reject or test that the test prefix is a valid PHP identifier:
```php
    public static function isIdentifier($id)
    {
        return 0 !== preg_match(
            '#^[a-zA-Z_\x7f-\xff][a-zA-Z0-9_\x7f-\xff]*$#',
            $id
        );
    }
```

remicollet:
Notice: this PR only plan to fix the test inconsistency, but yes, I agree, adding a test for correct identifier could make sense.

Hywan:
@remicollet You mean you prefer to see this PR fixed before checking the identifier?

remicollet:
> You mean you prefer to see this PR fixed before checking the identifier?

Yes. At least to get rid of spam about test failure by Koschei ;)

Hywan:
Would you be ok to do the next PR too?

remicollet:
>  Would you be ok to do the next PR too?

I can try (even if I'm not very familiar with atoum internals)

Hywan:
@remicollet Or at least open an issue :-).

Hywan:
I am merging your PR.

Hywan:
@remicollet How can we see the test report (or execution) on Koschei? How the error has been found?

Hywan:
@remicollet Since this is a bug fix, an immediate release has been created: https://github.com/atoum/atoum/releases/tag/2.2.1. Thanks!

remicollet:
@Hywan see link in PR description. On each build, you see what have change in the dependencies tree (build are launch when some dep. change, e.g. new PHP version, and also randomly), and a link to the build.log (which includes the test suite)

Hywan:
@remicollet Do you have a clue why the error happened on Fedora and not in Travis for instance?

remicollet:
@Hywan no, as said previously, very strange... my first idea was about the builder exhausting the random source. But I can't find any explanation reading the "uniqid" PHP function source... :(

Hywan:
Well, any way, it's good!

PR #492, opened 2015-08-27
Fix git pusher to correctly tag master (as dev-master) after a successful release

jubianchi:
cc @Hywan @mikaelrandy 

I chose `dev-master` to replace `DEVELOPMENT-master` as this is more consistent with what we use in the changelog and in composer.


Hywan:
:+1: for me.

jubianchi:
closed by 9f69f442f54f1497b4d8c764e3402bb22f10fc6f

mikaelrandy:
I see it too late (damn, holidays was good ;)) but it's :+1: for me too

PR #495, opened 2015-08-27
improve setTestMethodPrefix

remicollet:
As discussed in pr #491 , this is a try to improve setTestMethodPrefix by checking if arg is aither a valid PHP identifier or a valid regex.

Notice: I slightly improve isRegex method to really check if input is a valid regex (not seems to be a regex)

But don't know if this change really worth the work...

### DIFF
--- a/classes/test.php
+++ b/classes/test.php
@@ -1830,7 +1830,7 @@ abstract class test implements observable, \countable
 
 	private static function isRegex($namespace)
 	{
-		return is_numeric(@preg_match($namespace, 'foo'));
+		return false !== @preg_match($namespace, NULL);
 	}
 
 	private static function isValidIdentifier($id)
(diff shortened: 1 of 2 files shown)


classes/test.php
```diff
@@ -1826,6 +1830,14 @@ private static function cleanNamespace($namespace)
 
 	private static function isRegex($namespace)
 	{
-		return preg_match('/^([^\\\[:alnum:][:space:]]).*\1.*$/', $namespace) === 1;
+		return is_numeric(@preg_match($namespace, 'foo'));
```
jubianchi: Instead of hardcoding a random value, I would have used `null` to replace `' foo'`

jubianchi: Instead of `is_numeric` I would have tested the return value of `preg_match` against `false` which is what `preg_match` returns in case of error.

jubianchi:
I'm globally :+1: here.

Two comments.

remicollet:
Cosmetic changes for isRegex  done.

PR #496, opened 2015-08-31
Auto generate and inject mocks in test methods

jubianchi:
This PR allows users to define required mocks in test methods and get them auto-generated and injected by atoum. Here is an example:

```php
<?php

namespace tests\units;

use atoum;

class stdClass extends atoum {
    public function testFoo(\stdClass $a) {
        var_dump($a);
    }
}

```

This will produce the following output:

```
> PHP path: /usr/local/Cellar/php56/5.6.6/bin/php
> PHP version:
=> PHP 5.6.6 (cli) (built: May 14 2015 23:27:40)
=> Copyright (c) 1997-2015 The PHP Group
=> Zend Engine v2.6.0, Copyright (c) 1998-2015 Zend Technologies
=>     with Zend OPcache v7.0.4-dev, Copyright (c) 1999-2015, by Zend Technologies
=>     with Xdebug v2.3.2, Copyright (c) 2002-2015, by Derick Rethans
> tests\units\stdClass...
[0___________________________________________________________][1/1]
=> Test duration: 0.02 second.
=> Memory usage: 0.25 Mb.
> Total test duration: 0.02 second.
> Total test memory usage: 0.25 Mb.
> Running duration: 0.08 second.
Success (1 test, 0/1 method, 1 void method, 0 skipped method, 0 assertion)!
> There is 1 output:
=> In tests\units\stdClass::testFoo():
class mock\stdClass#67 (0) {
}
> There is 1 void method:
=> tests\units\stdClass::testFoo()
```

Everything works with type-hints. As soon as the type-hinted class has a constructor without any mandatory arguments, atoum will auto generate a mock from it. Otherwise it will throw an exception. 

Everything is done through data providers and this introduces an interesting side effect : atoum now supports closure data providers:

```php
<?php

namespace tests\units;

use atoum;

class stdClass extends atoum {
    public function beforeTestMethod($method) {
        $this->setDataProvider($method, function() {
            return array(array(new \stdClass()));
        });
    }

    public function testFoo(\stdClass $a) {
        var_dump($a);
    }
}
```

Finally, this PR will be able to evolve with PHP7 when scalar type hints will be supported we can imagine atoum will be able to generated random scalar values.

### DIFF
--- a/tests/units/classes/test.php
+++ b/tests/units/classes/test.php
@@ -965,9 +965,20 @@ namespace mageekguy\atoum\tests\units
 					->array($test->getDataProviders())->isEqualTo(array('testMethod1' => 'aDataProvider'))
 				->if($test = new dataProviderTest())
 				->then
-					->exception(function() use ($test, & $dataProvider) { $test->setDataProvider('testMethod2'); })
+					->object($test->setDataProvider('testMethod2'))->isIdenticalTo($test)
+					->array($providers = $test->getDataProviders())
+						->object['testMethod2']->isInstanceOf('closure')
+					->exception(function() use ($providers) { $providers['testMethod2'](); })
 						->isInstanceOf('mageekguy\atoum\exceptions\logic\invalidArgument')
-						->hasMessage('Could not generate a data provider for ' . get_class($test) . '::testMethod2() because SplFileInfo::__construct() has at least one mandatory argument')
+						->hasMessage('Could not generate a mock for ' . get_class($test) . '::testMethod2() because SplFileInfo::__construct() has at least one mandatory argument')
+				->if(
+					$test->getMockGenerator()->setDefaultNamespace('testMocks'),
+					$test->getMockGenerator()->allIsInterface()
+				)
+				->then
+					->array($providers['testMethod2']())->isEqualTo(array(array(new \testMocks\splFileInfo())))
+				->if($test = new dataProviderTest())
+				->then
 					->exception(function() use ($test, & $dataProvider) { $test->setDataProvider('testMethod3'); })
 						->isInstanceOf('mageekguy\atoum\exceptions\logic\invalidArgument')
 						->hasMessage('Could not generate a data provider for ' . get_class($test) . '::testMethod3() because it has at least one argument which is not type-hinted with a classname')
--- a/tests/units/classes/test.php
+++ b/tests/units/classes/test.php
@@ -987,6 +987,11 @@ namespace mageekguy\atoum\tests\units
 					->object($test->setDataProvider('testMethod1'))->isIdenticalTo($test)
 					->array($test->getDataProviders())
 						->object['testMethod1']->isInstanceOf('mageekguy\atoum\test\data\provider\aggregator')
+				->if($test = new dataProviderTest())
+				->then
+					->exception(function() use ($test, & $dataProvider) { $test->setDataProvider('testMethod3', function() {}); })
+						->isInstanceOf('mageekguy\atoum\exceptions\logic\invalidArgument')
+						->hasMessage('Cannot use a closure as a data provider for method ' . get_class($test) . '::testMethod3()')
 			;
 		}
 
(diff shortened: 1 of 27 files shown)


jubianchi:
From c08c293, if the user enable mock as interface (via `atoum\mock\generator::allIsInterface()`) atoum will auto-generate a mock even if the class constructor has mandatory arguments. In fact, mocking as interface won't mock the original class constructor, so this will be ok:

```php
<?php

namespace tests\units;

use atoum;

class bar {
    public function __construct($a) {}
}

class stdClass extends atoum {
    public function beforeTestMethod($method) {
        $this->getMockGenerator()->allIsInterface();
    }

    public function testFoo(\stdClass $a, bar $b) {
        var_dump($a);
    }
}
```

Mandragora2:
:+1:

vonglasow:
@jubianchi review ok for me

tests/units/classes/test.php
```diff
@@ -49,6 +49,18 @@ public function aDataProvider()
 		}
 	}
 
+	/**
+	@ignore on
```
mikaelrandy: invalid comment

mikaelrandy:
:+1:

jubianchi:
This seems to be all good: please test it and report problem so I can fix them and merge this feature!

Tiny example:

```php
<?php

namespace tests\units;

use atoum;

class stdClass extends atoum {
    public function testFoo(\stdClass $a) {
        $this->object($a);
    }

    public function testBar(\stdClass $a) {
        $this->string($a);
    }
}

```

Thanks!

Grummfy:
Remarque after test : 
* My test : Success (2 tests, 7/7 methods, 0 void method, 0 skipped method, 424 assertions)!
* Take more time (no bench mark), with 1 test with mock as args : from 1.06 to 1.26 second
* I use beforeTestMethod to fill mock and set the behaviour on it. It will be helpfull to have injection on this method too

jubianchi:
@Grummfy generating and injecting mocks in `beforeTestMethod` is doable but seems weird when we use it. You will have to do such things :

```php
<?php

namespace tests\units;

use atoum;

class stdClass extends atoum {
    protected $a;

    public function beforeTestMethod($method, \stdClass $a = null)
    {
        $this->a = $a;
    }
}

```
Your arguments will have to be nullable because the declaration of `beforeTestMethod` has to match the one in `atoum\test`.

Grummfy:
most of the time i do something like this : 

```php
<?php

namespace tests\units;

use atoum;

class stdClass extends atoum {
    protected $a;

    public function beforeTestMethod($method, \stdClass $a = null)
    {
        $a->getMockController()->stuff = function() {....}
        $this->a = $a;
    }
}
```
because I need common practice to int other stuff

jubianchi:
what seemed weird to me is declaring optionnal parameters when they are actually required for the test to run. It's not clear I think.

Grummfy:
Perhaps, but when I have more than half test method that require the same thing, I do this... perhaps I should use dataprovider  but beforeTestMEthod was easier to use ...

jubianchi:
What about something like this ?

```php
<?php

namespace tests\units;

use atoum;

class stdClass extends atoum {
    /** @provide \stdClass */
    protected function stdClassProvider()
    {
        $a = new \mock\stdClass;
        $a->getMockController()->stuff = function() {....}
        
        return $a;
    }

    public function testFoo(\stdClass $a) {
        $this->object($a);
    }
}
```

Grummfy:
I put the end of the comment from IRC here : 
> Grummfy : trop magique a mon goût ... mais j'aime bien quand même maintenant, du coup on est obliger d'avoir tout nos \stdClass qui passe par ce provider
> Grummfy : au final dans les tests que j'ai en tête
> Grummfy : c'est exactement comme cela que j'utilise mon beforeTestMethod
> Grummfy : question, si j'ai un namespace \A\B\C et que j'ai un use \A\B\C AS cc; je marque @provide cc ou \A\B\C ?
> Grummfy : si j'ai dans ma methode le namespace complet en input et l'alias est-ce que cela passe
> jubianchi : Grummfy: @provide \A\B\C
> Grummfy : toujours la classe de base jamais l'alias?
> Grummfy : idem pour l'automocking des argument des méthodes de test?
> jubianchi : Grummfy: pour les arguments, tu peux utiliser l'alias
> jubianchi : pour l'annotation, non
> jubianchi : enfin ça doit être faisable
> Grummfy : oki (faudra bien le préciser dans la doc ;))

Issue #498, opened 2015-09-11
generated .atoum.php config file does not help a new User


CedCannes:
I created a new project and I installed Atoum with composer.
I used the command `bin/Atoum --init` to generate configuration files.
I wanted the command `bin/atoum` enough to start my tests.
I think the file .atoum.php generated is incomplete and confusing to a new user.

Incomplete:
It should contain at least the line to add tests from a directory.
`$runner->addTestsFromDirectory('path/to/Tests/Units');`

Confusing:
I looked for a solution in the comments of the .atoum.php generated file to add tests from a directory and I wanted to try to do something that looked like what I was looking : 
```
$testGenerator = new atoum\test\generator();
$testGenerator->setTestClassesDirectory('path/to/Tests/Units');
$script->getRunner()->setTestGenerator($testGenerator);
```
But it has no effect. 
Error: No test found

mageekguy:
Test generator is related to… test generation ;), not test execution.
But i'm agree that we can improve the default `.atoum.php` generated by `--init`.
Please help all atoum's users (and their (very) busy developers) and do the PR ;) (FYI, the template used by `--init`is https://github.com/atoum/atoum/blob/master/resources/configurations/runner/atoum.php.dist).

mageekguy:
Oh sorry, just open my eyes and see your PR #499.
You can slap me with a wet trout!

PR #499, opened 2015-09-11
Update atoum.php.dist

CedCannes:
for #498


mageekguy:
I think that test generation is a feature less interesting than add all tests from a directory, so i think that you should move your comments before section related to test generation.
And thanks for your job!

Hywan:
:+1:

Hywan:
(even if a rebase would have been welcomed)

Issue #500, opened 2015-09-23
version of atoum in top of the command line


Grummfy:
Hello,
today when we launch the atoum command to test something we have something like this :
> PHP path: /usr/bin/php5
> PHP version:
=> PHP 5.6.4-4ubuntu6.2 (cli) (built: Jul  2 2015 15:29:28)
=> Copyright (c) 1997-2014 The PHP Group
=> Zend Engine v2.6.0, Copyright (c) 1998-2014 Zend Technologies
=>     with Zend OPcache v7.0.4-dev, Copyright (c) 1999-2014, by Zend Technologies
=>     with Xdebug v2.2.6, Copyright (c) 2002-2014, by Derick Rethans
> ...

What I want is something like this :
> PHP path: /usr/bin/php5
> PHP version:
=> PHP 5.6.4-4ubuntu6.2 (cli) (built: Jul  2 2015 15:29:28)
=> Copyright (c) 1997-2014 The PHP Group
=> Zend Engine v2.6.0, Copyright (c) 1998-2014 Zend Technologies
=>     with Zend OPcache v7.0.4-dev, Copyright (c) 1999-2014, by Zend Technologies
=>     with Xdebug v2.2.6, Copyright (c) 2002-2014, by Derick Rethans
> atoum version DEVELOPMENT-1.2.0 by Frédéric Hardy 
> atoum path: phar:///usr/share/atoum/mageekguy.atoum.phar
> ...

for example

jubianchi:
:+1:

Hywan:
:+1:

PR #501, opened 2015-09-23
Atoum fields cli report

jubianchi:
atoum_version_path


Grummfy:
thanks for fixing #500

jubianchi:
This should also be ported to hoaproject/Test#46

Hywan:
\o/

PR #502, opened 2015-09-25
Issue settestnamespace

remicollet:
Again, hit by uniqid returning "0"...  https://kojipkgs.fedoraproject.org/work/tasks/4560/11224560/build.log

1st commit is Like pr #491 
2nd is like pe #495 
2rd allow "foo\\bare" (don't know is this is expected)

### DIFF
--- a/classes/test.php
+++ b/classes/test.php
@@ -741,12 +741,18 @@ abstract class test implements observable, \countable
 
 	public function setTestNamespace($testNamespace)
 	{
-		$this->testNamespace = self::cleanNamespace($testNamespace);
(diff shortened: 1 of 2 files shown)


remicollet:
Patch applied in Fedora, build ok (will see if error occurs again... as this is really erratic)

jubianchi:
thanks @remicollet!

I did a quick review from my phone and everything seems ok. Will do another one from my laptop and merge if it ok.

Hywan:
@remicollet @jubianchi Seems correct for me too :+1:.

PR #505, opened 2015-09-30
Add travis configuration to deploy phar for tags

jubianchi:
This will allow us to automatically deploy the phar as a release artifact each time a tag is created.

You can see a sample green build here : https://travis-ci.org/jubianchi/atoum/jobs/82954506#L1505
And the associated release with the phar here : https://github.com/jubianchi/atoum/releases/tag/9.9.12


mikaelrandy:
:+1: 

Mais c'est normal l'API key versionnée ?

jubianchi:
@mikaelrandy it's encrypted so yes totally fine ;)

Hywan:
@jubianchi :+1: for me. I think this is not needed to include a line in the `CHANGELOG.md` for that. I updated the wiki consequently: https://github.com/atoum/atoum/wiki/How-to-release-atoum.

Hywan:
@jubianchi What about stubs? Should we open another issue/PR?

jubianchi:
@hywan stubs are released on their own and we produce no artifact for them. the only thing we should do there is to put a tag each time we release atoum

Issue #506, opened 2015-09-30
Access sub-n array through phpArray sugar asserter


mikaelrandy:
I have troubles to test sub-n arrays using phpArray sugar syntaxt `->array['key']`

When i test an array of first level with this syntax, all work fine. When i test various datas inside the array, all work fine. But when i want to test sub-n array (an array inside an array), it seems the asserter execute all test on first level array, not su-n level.

Here is an exemple : 
```php
<?php

namespace tests\units;

use atoum;

class StdClass extends atoum
{
    public function testExempleSugar()
    {
        $array = [
            'string'    => $string = uniqid(),
            'integer'   => $integer = rand(1, 1000),
            'object'    => ($object = new \StdClass()),
            'array'     => [
                'key1' => ($arrayVal1 = uniqid()), 
                'key2' => ($arrayVal2 = uniqid()),
            ],
        ];

        $this
            ->array($array)
                ->hasSize(4)
                ->string['string']
                    ->isEqualTo($string)
                ->integer['integer']
                    ->isEqualTo($integer)
                ->object['object']
                    ->isIdenticalTo($object)
                ->array['array']
                    ->hasSize(2)
                    ->hasKeys(['key1', 'key2'])
                    ->contains($arrayVal1)
                    ->contains($arrayVal2)
        ;
    }

    public function testExemple()
    {
        $array = [
            'string'    => $string = uniqid(),
            'integer'   => $integer = rand(1, 1000),
            'object'    => ($object = new \StdClass()),
            'array'     => [
                'key1' => ($arrayVal1 = uniqid()), 
                'key2' => ($arrayVal2 = uniqid()),
            ],
        ];

        $this
            ->array($array)
                ->hasSize(4)
                ->string($array['string'])
                    ->isEqualTo($string)
                ->integer($array['integer'])
                    ->isEqualTo($integer)
                ->object($array['object'])
                    ->isIdenticalTo($object)
                ->array($array['array'])
                    ->hasSize(2)
                    ->hasKeys(['key1', 'key2'])
                    ->contains($arrayVal1)
                    ->contains($arrayVal2)
        ;
    }
}
```

And the output
```shell
> PHP path: /usr/bin/php
> PHP version:
=> PHP 5.5.27 (cli) (built: Jul 23 2015 00:21:59)
=> Copyright (c) 1997-2015 The PHP Group
=> Zend Engine v2.5.0, Copyright (c) 1998-2015 Zend Technologies
[...]

PR #507, opened 2015-10-01
POC #506

GuillaumeDievart:


GuillaumeDievart:
Don't merge it, it's only to have feedback, I have only test the method hasSize.

jubianchi:
@GuillaumeDievart finally I remember what the real problem is with this. For example:

```php
$this
    ->array($foo)
        ->hasSize(3)
        ->string['bar']->isEqualTo('bar')
        ->array['baz']
            ->hasSize(1)
            ->string['baz']->isEqualTo('baz')
        ->array['boo']
            ->hasSize(1)
            ->string['boo']->isEqualTo('boo')
;
```

Take a look at the indentation because this express what I mean in my test : I want to assert on the `$foo` array, then on `$foo['baz']` array and finally on `$foo['boo']`. The problem is we can't tell atoum that at some point we want to get back to the toplevel array. 

This is the real problem of allowing nested `array` asserters.

GuillaumeDievart:
Ok, I see ! We could imagine to add a method fromParent, or it's really ugly ?

mikaelrandy:
Is it imaginable to create a new asserter to explicitly tell to go back to parent level ?
Like the `->end()` operator in [Symfony Configuration](http://symfony.com/doc/current/components/config/definition.html#variable-nodes).

So, the @jubianchi example could become : 
```php
$this
    ->array($foo)
        ->hasSize(3)
        ->string['bar']->isEqualTo('bar')
        ->array['baz']
            ->hasSize(1)
            ->string['baz']->isEqualTo('baz')
        ->end()
        ->array['boo']
            ->hasSize(1)
            ->string['boo']->isEqualTo('boo')
;
```

PS : if someone has a better name than this ugly "end", i will be the first to be interested ;)

jubianchi:
@GuillaumeDievart @mikaelrandy we always tried to avoid such things.

We have to be imaginative here to find something to avoid create a `end` (or whatever)  method. perhaps something like:

```php
$this
    ->array($foo)
        ->hasSize(3)
        ->string['bar']->isEqualTo('bar')
        ->array['baz']
            ->hasSize(1)
            ->string['baz']['baz']->isEqualTo('baz')
        ->array['boo']
            ->hasSize(1)
            ->string['boo']['boo']->isEqualTo('boo')
;
```

So the `array` asserter will always target the toplevel array. But we could use nested keys to target child values. HTe only thing to fix then will be this:

```php
$this
    ->array($foo)
    ->array['baz']
        ->hasSize(1)
```

The `hasSize(1)` has to match against the child array not the parent.

mikaelrandy:
@jubianchi i understand the point and i totally agree with.
But i think your workaround is not compatible with the atoum fluent interface :'( 

Clearly, i would be less disappointed if there is not "array" sugar asserter inside phpArray asserter than this way to transform all other sugar asserters.

My 2cts

jubianchi:
@mikaelrandy you can already bypass the sugar-syntax to access nested arrays:

```php
$this
    ->array($array)
        ->hasSize(4)
        ->hasKeys(['string', 'integer', 'object', 'array'])
        ->string['string']
            ->isEqualTo($string)
        ->integer['integer']
            ->isEqualTo($integer)
        ->object['object']
            ->isIdenticalTo($object)
        ->array($array['array'])
            ->hasSize(2)
            ->hasKeys(['key1', 'key2'])
            ->contains($arrayVal1)
            ->contains($arrayVal2)
;
```

mikaelrandy:
@jubianchi it's exactly the workaround that i've used to continue to work ;)

GuillaumeDievart:
Hi,

I can close this PR ?

jubianchi:
@GuillaumeDievart I think we will keep your PR open and make it evolve a bit. @agallou came with a pretty nice idea: why not use something like symfony property access ?

```php
$this
    ->array($foo)
        ->hasSize(3)
        ->string['bar']->isEqualTo('bar')
        ->array['baz']->hasSize(1)
            ->string['baz.baz']->isEqualTo('baz')
        ->array['boo']->hasSize(1)
            ->string['boo.boo']->isEqualTo('boo')
;
```

jubianchi:
I managed to get something working with @agallou idea:

```php
<?php

namespace tests\units;

class stdClass extends \mageekguy\atoum
{
    public function testFoo()
    {
        $actual = [
            'foo',
            'bar',
            [
                'baz',
                'boo',
                'baa',
                [],
                '4.1' => 'ho'
            ],
            '3.0' => 'hey'
        ];

        $this
            ->array($actual)
                ->hasSize(4)
                ->string[0]->isEqualTo('foo')
                ->string[1]->isEqualTo('bar')
                ->string['2.1']->isEqualTo('boo')
                ->string['3.0']->isEqualTo('hey')
                ->array[2]
                    ->hasSize(5)
                    ->string['2.0']->isEqualTo('baz')
                    ->array['2.3']->isEmpty
                    ->string['2.4.1']->isEqualTo('ho')

                //Example failures:
                //
                //->string['2.3.0']->isEqualTo('something')     // array(4) has no key string(1) '3'
                //->string['2.4.0']->isEqualTo('something')     // array(5) has no key string(3) '4.0'
                //->string['0.qsdqd']->isEqualTo('something')   // Value string(3) 'foo' at key 0 is not an array
                //->string['2.1.qsdqd']->isEqualTo('something') // Value string(3) 'bar' at key 1 is not an array
        ;
    }
}
```

@atoum/all let me know what you think about that.

marmotz:
And if key contains dot?

jubianchi:
@marmotz it handles dot in keys as you can see in the example but there is still some case which are hard to desambiguate this is why I came up with something a bit different:

```
<?php

namespace tests\units;

class stdClass extends \mageekguy\atoum
{
    public function testFoo()
    {
        $actual = [
            0 => 'foo',
            1 => 'bar',
            2 => [
                3 => 'baz',
                4 => 'boo',
                5 => 1,
                6 => [],
                7 => new \stdClass,
                '4.1' => 'ho'
            ],
            '3.0' => 'hey'
        ];

        $this
            ->array($actual)->hasSize(4)
                ->string[0]->isEqualTo('foo')
                ->string['3.0']->isEqualTo('hey')
                ->string[$this->array->key(1)]->isEqualTo('bar')

                ->array[2]->hasSize(6)
                    ->string[$this->array->key(2, 3)]->isEqualTo('baz')
                    ->string[$this->array->key(2, 4)]->isEqualTo('boo')
                    ->integer[$this->array->key(2, 5)]->isEqualTo(1)
                    ->array[$this->array->key(2, 6)]->isEmpty
                    ->object[$this->array->key(2, 7)]->isInstanceOf(\stdClass::class)
                    ->string[$this->array->key(2, '4.1')]->isEqualTo('ho')

                //Example failures:
                //
                //->string[$this->array->key(2, 6, 0)]->isEqualTo('something')     // failed: array(0) has no key integer(0)
                //->string[$this->array->key(2, '4.0')]->isEqualTo('something')     // array(6) has no key string(3) '4.0'
                //->string[$this->array->key(0, 'qsdqd')]->isEqualTo('something')   // Value string(3) 'foo' at key 0 is not an array
                //->string[$this->array->key(2, 3, 'qsdqd')]->isEqualTo('something') // Value string(3) 'baz' at key 3 is not an array
        ;
    }
}

```

Grummfy:
Second example is hard to read.
Otherwise it's :+1:

marmotz:
I think it's not readable at all. And what I love in atoum is its readability.

What about this kind of code :
```php
        $this
            ->array($actual)->hasSize(4)
                ->string(0)->isEqualTo('foo')
                ->string('3.0')->isEqualTo('hey')
                ->string(1)->isEqualTo('bar')
```

But I think it's impossible to make difference between `asserter($variable)` and `asserter(...$keys)`.

Maybe we should add a keyword `->children()` to mark all future asserter as children asserter ? (And maybe a `->endChildren()`)

```php
        $this
            ->array($actual)->hasSize(4)
                ->children()
                    ->string(0)->isEqualTo('foo')
                    ->string('3.0')->isEqualTo('hey')
                    ->string(1)->isEqualTo('bar')
                ->endChildren()
            ->string($anOtherStringToTest)
```

Or we should use closures ?

```php
        $this
            ->array($actual)->hasSize(4)
                ->children(
                    function($children) {
                        $children
                            ->string(0)->isEqualTo('foo')
                            ->string('3.0')->isEqualTo('hey')
                            ->string(1)->isEqualTo('bar')
                        ;
                     }
                 )
```

What do you think ?

jubianchi:
@marmotz :+1: for the closure, nice idea!

mikaelrandy:
in my opinion, @marmotz first proposition is more readable than the one with the close

jubianchi:
@mikaelrandy wich one ? the one with the `->endChildren()` ?

As I told earlier in this issue, we always avoided such things in atoum because they introduce noise in tests and they are really hard to implement: we'll have to maintain an asserter stack inside the array asserter and push/pop in it as the user navigates.

The proposed solution with the closure is pretty simple to implement and just does the job well without too much noise. Also, closures are closer to everything we do in atoum with others asserters (`when`, `exception`, ...).

The solution I proposed with the helper function (`$this->array->key`) is already working but I admit that it's not the most readable. We can make it a bit simpler by directly exposing the helper as a plain old function:

```
<?php

namespace tests\units;

// PHP >= 5.6
use function mageekguy\atoum\helpers\key;

// PHP < 5.6
use mageekguy\atoum\helpers

class stdClass extends \mageekguy\atoum
{
    public function testFoo()
    {
        $actual = [
            0 => 'foo',
            1 => 'bar',
            2 => [
                3 => 'baz',
                4 => 'boo',
                5 => 1,
                6 => [],
                7 => new \stdClass,
                '4.1' => 'ho'
            ],
            '3.0' => 'hey'
        ];

        $this
            ->array($actual)->hasSize(4)
                ->string[key(2, 3)]->isEqualTo('baz') // PHP >= 5.6
                ->string[$this->array->key(2, 4)]->isEqualTo('boo') // PHP < 5.6
                ->integer[helpers\key(2, 5)]->isEqualTo(1) // PHP < 5.6
        ;
    }
}
```

mikaelrandy:
@jubianchi yeah, i remember ;)

By stepping back and considering only readability, i find the first proposal more readable ;)
But the argument about atoum already use closure for other stuff is a very good argument !

jubianchi:
@mikaelrandy @marmotz @GuillaumeDievart what if I tell you we don't need any closure or weird things like I proposed before ?

See:

```php
<?php

namespace tests\units;

class stdClass extends \mageekguy\atoum
{
    public function testExempleSugar()
    {
        $array = [
            'string'    => $string = uniqid(),
            'integer'   => $integer = rand(1, 1000),
            'object'    => $object = new \StdClass,
            'array'     => [
                'key1' => $arrayVal1 = uniqid(),
                'key2' => $arrayVal2 = uniqid(),
                'key3' => []
            ],
        ];

        $this
            ->array($array)
                ->hasSize(4)
                ->string['string']
                    ->isEqualTo($string)
                ->integer['integer']
                    ->isEqualTo($integer)
                ->array['array']
                    ->hasSize(3)
                    ->hasKeys(['key1', 'key2'])
                    ->contains($arrayVal1)
                    ->string['array']['key2']->isEqualTo($arrayVal2)
                    ->array['array']['key3']->isEmpty
                ->object['object']
                    ->isIdenticalTo($object)
        ;
    }
}
```

This is the test @mikaelrandy dropped in #507 and with a minor patch in the `phpArray` asserter it passes ;)

GuillaumeDievart:
:+1 for the closure

marmotz:
For once an idea from me please...

mikaelrandy:
@jubianchi i drop a test ?

Anyway, for me, your proposal seems to be harder to read, but lighter.
I don't have a clear favorite :/

jubianchi:
Here is the final proposal I can make based on @marmotz's idea (with closure):

```php
<?php

namespace tests\units;

class stdClass extends \mageekguy\atoum
{
    public function testExempleSugar()
    {
        $array = [
            'string'    => $string = 'foo',
            'integer'   => $integer = rand(1, 1000),
            'object'    => $object = new \StdClass,
            'array'     => [
                'key1' => $arrayVal1 = uniqid(),
                'key2' => $arrayVal2 = uniqid(),
                'key3' => []
            ],
        ];

        $this
            ->array($array)->hasSize(4)
                ->string['string']->isEqualTo($string)
                ->integer['integer']->isEqualTo($integer)
                ->child['array'](function($child) use ($arrayVal1, $arrayVal2) {
                        $child->hasSize(3)
                            ->hasKeys(['key1', 'key2'])
                            ->contains($arrayVal1)
                            ->string['key2']->isEqualTo($arrayVal2)
                            ->child['key3'](function($child) { $child->isEmpty; })
                        ;
                    }
                )
                ->string['array']['key1']->isEqualTo($arrayVal1)
                ->object['object']
                    ->isIdenticalTo($object)
        ;
    }
}
```

marmotz:
I LIKE (to move) IT !

Hywan:
Much better :+1:.

Hywan:
@jubianchi What about `->child('keyname', function () { … })` instead of `->child['keyname'](function () { … })`, it sounds less tricky.

Hywan:
Also `->child[…](…)` is not necessarily compatible with PHP5.3 (which is still supported by us).

marmotz:
It seems that Hywan proposal is more readable

mikaelrandy:
I like the @Hywan proposal !

jubianchi:
@Hywan proposal is a bit more readable but not consistent with other array assertions.

And it works on 5.3

Hywan:
@jubianchi This is OK for me to have a difference syntax for nodes and leafes when traversing a structure. And readability is more important I guess :-/.

jubianchi:
The readability here is a detail, there is no big difference between

```->child['key3'](function($child) { $child->isEmpty; })``` and
```->child('key3', function($child) { $child->isEmpty; })```

```diff
diff --git i/a.php w/a.php
index 4e2ba3e..77bbfbe 100644
--- i/a.php
+++ w/a.php
@@ -1 +1 @@
->child[-[-]{+(+}'key3'[-](-]{+, +}function() { ->isEmpty; })
```

This is a two-characters diff...

To me, consistency is more important : in atoum, every function-asserter assert on the value passed as their first argument. With what you propose it won't be the case. In the other hand, every array-key-asserter take the key as if we were accessing an array. It makes sense to me to keep this.

jubianchi:
Closing this one in favor of #540

PR #509, opened 2015-10-01
suggest atoum/stubs ( #508 )

### LINKED ISSUE #508: Suggest atoum/stubs for dependency?
I think it could be interesting to suggest in composer to add atoum/stubs.

What's your advice on this ?

Grummfy:
fix #508

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -51,7 +51,7 @@
 	},
 	"suggest": {
 		"ext-mbstring": "Provides support for UTF-8 strings",
-		"atoum/stubs": "Provides stubs for atoum that will help ide for autocompletion"
+		"atoum/stubs": "Provides IDE support (like autocompletion) for atoum"
 	},
 	"extra": {
 		"branch-alias": {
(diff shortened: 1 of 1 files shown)


composer.json
```diff
@@ -50,7 +50,8 @@
 		"classmap": [ "classes/" ]
 	},
 	"suggest": {
-		"ext-mbstring": "Provides support for UTF-8 strings"
+		"ext-mbstring": "Provides support for UTF-8 strings",
+		"atoum/stubs": "Provides stubs for atoum that will help ide for autocompletion"
```
Hywan:
```diff
- ide
+ IDE
```
Actually, the whole sentence is not very clearn.
What about:
> Provides IDE support (like autocompletion)

This is simple but it's understandable.

jubianchi: +1 with @Hywan

Grummfy:
How can I improve this?

composer.json
mikaelrandy: :+1: too

jubianchi:
@Grummfy just change the wording and it should be good ;) @Hywan's suggestion seems good to me.

jubianchi:
oh yeah! :+1:

Hywan:
:+1:

PR #515, opened 2015-10-28
Fix PHP7 support in the basic resolver

Hywan:
Since PHP7, some new keywords have been introduced, like `string` or `float` for instance. Consequently, `asserters/string.php` have been renamed to `asserters/phpString.php` etc.

However, it is like we forget to change the resolver too. This bug was not visible because asserter aliases exist in the famous `test` class. Nevertheless, if we use only the `asserter\generator`, the bug is present.

This patch fixes the bug by checking twice if the class exists in the resolver: First time with the canonical asserter name, second time prefixed by `php` and with the first letter uppercased.

### DIFF
--- a/classes/asserter/resolver.php
+++ b/classes/asserter/resolver.php
@@ -2,6 +2,8 @@
 
 namespace mageekguy\atoum\asserter;
 
+use mageekguy\atoum\test;
+
 class resolver
 {
 	const defaultBaseClass = 'mageekguy\atoum\asserter';
@@ -50,13 +52,28 @@ class resolver
 		{
 			$class = $this->checkClass($asserter);
 		}
-		else foreach ($this->namespaces as $namespace)
+		else
 		{
-			$class = $this->checkClass($namespace . '\\' . $asserter);
-
-			if ($class !== null)
+			if (false === test::isValidIdentifier($asserter))
 			{
-				break;
+				return null;
+			}
+
+			foreach ($this->namespaces as $namespace)
+			{
+				$class = $this->checkClass($namespace . '\\' . $asserter);
+
+				if ($class !== null)
+				{
+					break;
+				}
+
+				$class = $this->checkClass($namespace . '\\php' . ucfirst($asserter));
+
+				if ($class !== null)
+				{
+					break;
+				}
 			}
 		}
 
--- a/classes/asserter/resolver.php
+++ b/classes/asserter/resolver.php
@@ -49,6 +49,11 @@ class resolver
 
 	public function resolve($asserter)
 	{
+		if (false === analyzer::isValidNamespace($asserter))
+		{
+			return null;
+		}
+
 		$class = null;
 
 		if (strpos($asserter, '\\') !== false)
@@ -57,11 +62,6 @@ class resolver
 		}
 		else
 		{
-			if (false === analyzer::isValidIdentifier($asserter))
-			{
-				return null;
-			}
-
 			foreach ($this->namespaces as $namespace)
 			{
 				$class = $this->checkClass($namespace . '\\' . $asserter);
--- a/classes/asserter/resolver.php
+++ b/classes/asserter/resolver.php
@@ -2,7 +2,10 @@
 
 namespace mageekguy\atoum\asserter;
 
-use mageekguy\atoum\test;
+use
+	mageekguy\atoum\test,
+	mageekguy\atoum\tools\variable\analyzer
+;
 
 class resolver
 {
@@ -54,7 +57,7 @@ class resolver
 		}
 		else
 		{
-			if (false === test::isValidIdentifier($asserter))
+			if (false === analyzer::isValidIdentifier($asserter))
 			{
 				return null;
 			}
--- a/classes/test.php
+++ b/classes/test.php
@@ -749,7 +749,7 @@ abstract class test implements observable, \countable
 			throw new exceptions\logic\invalidArgument('Test namespace must not be empty');
 		}
 
-		if (!analyzer::isRegex($testNamespace) && !analyzer::isValidNamespace($testNamespace, true))
+		if (!analyzer::isRegex($testNamespace) && !analyzer::isValidNamespace($testNamespace))
 		{
 			throw new exceptions\logic\invalidArgument('Test namespace must be a valid regex or identifier');
 		}
--- a/classes/test.php
+++ b/classes/test.php
@@ -1841,7 +1841,7 @@ abstract class test implements observable, \countable
 		return false !== @preg_match($namespace, null);
 	}
 
-	private static function isValidIdentifier($identifier)
+	public static function isValidIdentifier($identifier)
 	{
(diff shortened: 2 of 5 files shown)


classes/asserter/resolver.php
```diff
@@ -50,13 +52,28 @@ public function resolve($asserter)
 		{
 			$class = $this->checkClass($asserter);
 		}
-		else foreach ($this->namespaces as $namespace)
+		else
 		{
-			$class = $this->checkClass($namespace . '\\' . $asserter);
+			if (false === test::isValidIdentifier($asserter))
+			{
+				return null;
```
jubianchi: I would rather throw an `invalidArgument` exception here.

classes/asserter/resolver.php
```diff
@@ -50,13 +52,28 @@ public function resolve($asserter)
 		{
 			$class = $this->checkClass($asserter);
 		}
-		else foreach ($this->namespaces as $namespace)
+		else
 		{
-			$class = $this->checkClass($namespace . '\\' . $asserter);
+			if (false === test::isValidIdentifier($asserter))
```
jubianchi: Adding a hard dependency on the `test` class here does not seem clean.

I would rather duplicate the `isValidIdentifier` method in the `resolver` class for now and see how we can extract this in a dedicated class later.

jubianchi: or add thos check methods `isValid*` here : https://github.com/atoum/atoum/blob/master/classes/tools/variable/analyzer.php

tests/units/classes/asserter/resolver.php
```diff
@@ -54,10 +54,10 @@ public function testResolve()
 					$this->function->is_subclass_of = true
 				)
 				->then
-					->string($this->testedInstance->resolve($asserter = uniqid()))->isEqualTo('mageekguy\atoum\asserters\\' . $asserter)
-					->string($this->testedInstance->resolve($asserter = '\\' . uniqid()))->isEqualTo($asserter)
-					->string($this->testedInstance->resolve($asserter = uniqid() . '\\' . uniqid()))->isEqualTo($asserter)
-					->string($this->testedInstance->resolve($asserter = '\\' . uniqid() . '\\' . uniqid()))->isEqualTo($asserter)
+					->string($this->testedInstance->resolve($asserter = 'a' . uniqid()))->isEqualTo('mageekguy\atoum\asserters\\' . $asserter)
```
jubianchi: please use the `$prefix` argument of `uniqid`

jubianchi:
changelog please :)

classes/asserter/resolver.php
Hywan: I kept the same behavior. Do you confirm we would like to change the behavior?

jubianchi: AFAIR we did not run this kind of check before but you are right, with the actual code (before your patch) if something was wrong (asserter not found) we just returned `null`. I'm wondering if this is good.

It does not make sense to return `null` for an invalid identifier: it's not valid so there will never be any asserter class matching it. Requesting an asserter with an invalid identifier is just wrong IMHO.

But now I have another question: why not moving the `isValidIdentifier` to the very beginning of the `resolve` method ?

classes/asserter/resolver.php
Hywan: :white_check_mark:

classes/asserter/resolver.php
Hywan: Because, I don't know why, we can pass a `\\` as the asserter name. In this case, validation will fail. So in order to keep the same behavior… (same story ;-)).

Hywan:
Done.

classes/asserter/resolver.php
```diff
@@ -50,13 +55,28 @@ public function resolve($asserter)
 		{
 			$class = $this->checkClass($asserter);
 		}
-		else foreach ($this->namespaces as $namespace)
+		else
 		{
-			$class = $this->checkClass($namespace . '\\' . $asserter);
+			if (false === analyzer::isValidIdentifier($asserter))
```
jubianchi: please can you move this check to the beginning of the method ?

classes/test.php
```diff
@@ -748,7 +749,7 @@ public function setTestNamespace($testNamespace)
 			throw new exceptions\logic\invalidArgument('Test namespace must not be empty');
 		}
 
-		if (!self::isRegex($testNamespace) && !self::isValidNamespace($testNamespace, true))
+		if (!analyzer::isRegex($testNamespace) && !analyzer::isValidNamespace($testNamespace, true))
```
jubianchi: Using static calls here does not seem to be good: I think this is why we need to mock the `preg_match` function in some tests.

Using an instance of `analyzer` and injecting it in the `test` instance might be better for testability.

classes/asserter/resolver.php
Hywan: Did you see https://github.com/atoum/atoum/pull/515#discussion_r43242078?

classes/asserter/resolver.php
jubianchi: We can but it's not valid ;)

What we can do is ask for `\\SomeAsserter` and this is what we try to check in the first `if`

classes/test.php
Hywan: Should I create the instance in the constructor of `test`?

classes/asserter/resolver.php
Hywan: This is more complex actually. We can have a FQCN, like `my\asserters\foo`. So I will split the string by `\` and check each part if valids.

Hywan: We have `isValidNamespace`, so here we go.

Hywan:
@jubianchi Only https://github.com/atoum/atoum/pull/515#discussion_r43248633 is pending.

classes/asserter/resolver.php
Hywan: :white_check_mark:

classes/tools/variable/analyzer.php
```diff
@@ -82,4 +82,25 @@ public function isArray($mixed)
...
+	}
+
+	public static function isValidIdentifier($identifier)
+	{
+		return 0 !== \preg_match('#^[a-zA-Z_\x7f-\xff][a-zA-Z0-9_\x7f-\xff]*$#', $identifier);
+	}
+
+	public static function isValidNamespace($namespace)
+	{
+		foreach(explode('\\', trim($namespace, '\\')) as $sub) {
```
jubianchi: `{`should be on its own line

classes/tools/variable/analyzer.php
```diff
@@ -82,4 +82,25 @@ public function isArray($mixed)
...
+			if (!self::isValidIdentifier($sub)) {
```
jubianchi: `{`should be on its own line

Hywan: :white_check_mark:

classes/tools/variable/analyzer.php
Hywan: :white_check_mark:

Hywan:
Ready?

Hywan:
Thanks!

PR #516, opened 2015-11-04
Add a "callStaticOnTestedClass" method

### LINKED ISSUE #514: Add a callStaticOnTestedInstance() function
Atoum proposes [some very helpfully function around testedInstance](http://docs.atoum.org/en/latest/written_help.html#newtestedinstance-testedinstance).

But, to execute static method on tested class, or call constant on it is difficult:
```php
$this
    ->given($testedClass = $this->testedClass->getClass())
        ->then
            ->variable($testedClass::myStaticMethod())
```

or
```php
$this
    ->then
        ->variable(call_user_func([ $this->getTestedClassName(), 'myStaticMethod’]))
```

It should be very helpful to have a `callStaticOnTestedClass('myStaticMethod')` or `getConstantOnTestedClass('constName')`

mikaelrandy:
Allow to easilly call static method on testedClass.

Close #514

### DIFF
--- a/classes/test.php
+++ b/classes/test.php
@@ -420,6 +420,17 @@ abstract class test implements observable, \countable
 			->setPropertyHandler('exception', function() { return asserters\exception::getLastValue(); })
 		;
 
+		// Around testedInstance
+		$this->assertionManager
+			->setHandler('callStaticOnTestedClass', function() use ($test) {
+					$args = func_get_args();
+					$method = array_shift($args);
+
+					return call_user_func_array(array($test->getTestedClassName(), $method), $args);
+				}
+			)
+		;
+
 		$mockGenerator = $this->mockGenerator;
 
 		$this->assertionManager
--- a/tests/units/classes/test.php
+++ b/tests/units/classes/test.php
@@ -4,6 +4,19 @@ namespace mageekguy\atoum
 {
 	class emptyTest {}
 	class notEmptyTest {}
+
+	class withStatic
+	{
+		static public function staticMethod($return)
+		{
+			return $return;
+		}
+
+		static public function someOtherStaticMethod($return1, $return2, $return3)
+		{
+			return array($return1, $return2, $return3);
+		}
+	}
 }
 
 namespace mageekguy\atoum\mock\mageekguy\atoum
@@ -56,6 +69,26 @@ namespace mageekguy\atoum\tests\units
 			$this->setTestedClassName('mageekguy\atoum\test');
 
 			parent::__construct();
+		}
+	}
+
+	class withStatic extends atoum\test
+	{
+		public function testStaticMethod()
+		{
+			$this
+				->then
+					->string($this->callStaticOnTestedClass('staticMethod', $return = uniqid()))
+						->isEqualTo($return)
+
+					->array($this->callStaticOnTestedClass(
+						'someOtherStaticMethod',
+						$return1 = uniqid(),
+						$return2 = uniqid(),
+						$return3 = uniqid()
+					))
+						->isEqualTo(array($return1, $return2, $return3))
+			;
 		}
 	}
 
--- a/tests/units/classes/test.php
+++ b/tests/units/classes/test.php
@@ -34,15 +34,15 @@ namespace mageekguy\atoum\tests\units
 	require_once __DIR__ . '/../runner.php';
 
 	/**
-	@ignore on
-	@tags empty fake dummy
-	@maxChildrenNumber 666
-	*/
+	 * @ignore on
+	 * @tags empty fake dummy
+	 * @maxChildrenNumber 666
+	 */
 	class emptyTest extends atoum\test {}
 
 	/**
-	@ignore on
-	*/
+	 * @ignore on
+	 */
 	class notEmptyTest extends atoum\test
 	{
 		/**
@@ -62,6 +62,9 @@ namespace mageekguy\atoum\tests\units
 		}
 	}
 
+	/**
+	 * @ignore on
+	 */
 	class foo extends atoum\test
 	{
 		public function __construct()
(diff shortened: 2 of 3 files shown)


mikaelrandy:
Code pushed at WIP status to add some help (ping @jubianchi)

classes/test.php
```diff
@@ -1181,6 +1181,12 @@ public function runTestMethod($testMethod, array $tags = array())
 						)
 					;
 
+					$this->factoryBuilder->build($testedClass)
+						->addToAssertionManager($this->assertionManager, 'callStaticOnTestedInstance', function($methodName) use ($testedClass) {
```
jubianchi: you don't need to use the factory builder in your case ;)

You should rather use something like : https://github.com/atoum/atoum/blob/master/classes/test.php#L371

```php
$this->assertionManager
    ->setHandler('callStaticOnTestedInstance', function () use ($test) {
            $args = func_get_args();
            $method = array_shift($args);

            return call_user_func_array(array($test->getTestedCLassName(), $method), $args);
        }
    )
; 
```

tests/units/classes/test.php
```diff
@@ -57,6 +57,16 @@ public function __construct()
...
 		}
+
+		static public function staticMethod($return)
+		{
+			return $return;
+		}
+
+		static public function someOtherStaticMethod($return1, $return2, $return3)
+		{
+			return [$return1, $return2, $return3];
```
jubianchi: be careful with short-array syntax, we are old-school guys and still support php 5.3 ;)

classes/test.php
jubianchi: @mikaelrandy the factory builder is used to build a factory which will help atoum instanciate tested class.

Because we don't know in advance how to instanciate the tested class, we had to write a builder to dynamically create factories. 

This is the only goal of the factory builder ;)

In your case, you don't have to instanciate the tested class because you just want to use static calls.

_related to your question on IRC_

tests/units/classes/test.php
```diff
@@ -466,6 +479,22 @@ public function testGetTestedClassPath()
 			;
 		}
 
+		public function testCallStaticOnTestedClass()
+		{
+			$this
+				->if($this->setTestedClassName('\mageekguy\atoum\withStatic'))
+				->then
+					->string($this->callStaticOnTestedClass('staticMethod', $return = uniqid()))->isEqualTo($return)
```
mikaelrandy: This test failed cause the `setTestedClassName` raise an (normal) exception.

Any clue to how i can test my new asserter ? ping @jubianchi @mageekguy

jubianchi: The `atoum\test` class is one of the most complicated class to test. Even if your feature looks quite simple at first, writing tests for it won't be as easy as we might think. There are some gotchas you have to avoid:

* The first (and the one you encounter) is you can't use `$this` as your tested instance in this test. The `$this`  represents the test being actually ran by atoum. You can't alter this one. You have to manually instanciate a test instance for your test case and work with it. Look at this line for an example : https://github.com/mikaelrandy/atoum/blob/feature/callStaticOnTestedClass/tests/units/classes/test.php#L174
* To be able to instanciate you own tested test, you will have to write a fake test class which you already did in your PR : https://github.com/mikaelrandy/atoum/blob/feature/callStaticOnTestedClass/tests/units/classes/test.php#L8
* Last but not least: creating a bug in the test class might fake your test suite result or avoid it from running correctly.

mikaelrandy: Thanks for the answer @jubianchi 
It seems strange to me to do a fake test to test the test tool itself ôO

But i do it.

Thanks ;)

mikaelrandy:
All seems to be ok now ;)

jubianchi:
Once you merge [my PR](https://github.com/mikaelrandy/atoum/pull/2) on your branch, this one will be ready.

I'll let you merge on your branch and then click the green button here if everything if fine on travis ;)

thanks for your patch ;)

mikaelrandy:
@jubianchi i have made a mistake while rebasing on master and i have lost the merge of your PR.
But i have rebase the commit manually (with a little update)

classes/test.php
```diff
@@ -7,7 +7,7 @@
 	mageekguy\atoum\test,
 	mageekguy\atoum\mock,
 	mageekguy\atoum\asserter,
-	mageekguy\atoum\asserters,
+	mageekguy\atoum\assercters,
```
jubianchi: there is a tiny typo here :)

classes/test.php
```diff
@@ -396,6 +396,16 @@ public function setAssertionManager(test\assertion\manager $assertionManager = n
 			->setPropertyHandler('exception', function() { return asserters\exception::getLastValue(); })
 		;
 
+		// Around testedInstance
+		$this->assertionManager
+			->setHandler('callStaticOnTestedClass', function() use ($test) {
```
jubianchi: @mikaelrandy can you change indentation a bit here please :

```php
$this->assertionManager
			->setHandler('callStaticOnTestedClass', function() use ($test) {
					//...
				}
			)
;
```

jubianchi:
@mikaelrandy I saw it ;) no problem for the lost commit, don't worry.

The diff seems good, just to minor issues ;)

classes/test.php
mikaelrandy: It's an indentation using tab, like the rest of the file.
Do you want to change the indentation just for this addition ?

jubianchi: @mikaelrandy what is not correct is the indentation inside the closure : we always add one extra level for nested closures ;) See the closing brace (`}`)  and the closing paren. (`)`) ?

mikaelrandy: Oh ! I see, sorry for my mistake ;)

PR #520, opened 2015-11-10
Introduce the constant mocker

### LINKED ISSUE #511: Mock constant
Hello :-),

We have `$this->function->$functionName = function () { … }` to mock a function easily. Is it possible to use the same mechanism to mock a constant? I tried `$this->constant->$constantName` without no success. Do you have any clue?

Thanks!

cc @mageekguy, @jubianchi

Hywan:
Fix #511.

In 2 steps

### Split PHP mocker

The PHP mocker is splitted into 2 layers:
  1. Abstract mocker, containing all namespace logic,
  2. Function mocker.

The goal is to be able to introduce another mocker. So we must decouple this part first.

**BC break**: All “phpMocker” references become “phpFunctionMocker” (like `setPhpMocker`, `getPhpMocker` etc.). This is not complicated to migrate but this is necessary, sorry.

### Introduce the constant mocker

We use it the same way we use the function mocker:
```php
$this->constant->PHP_VERSION_ID = '606060'; // troll \o/
```

Note: This is used to mock constants, not class constants.

Thoughts?

### DIFF
--- a/classes/php/mocker/funktion.php
+++ b/classes/php/mocker/funktion.php
@@ -1,0 +1,184 @@
+<?php
+
+namespace mageekguy\atoum\php\mocker;
+
+use
+	mageekguy\atoum\exceptions,
+	mageekguy\atoum\php\mocker
+;
+
+class funktion extends mocker
+{
+	public function __construct($defaultNamespace = '')
+	{
+		parent::__construct($defaultNamespace);
+		$this->setReflectedFunctionFactory();
+	}
+
+	public function __get($functionName)
+	{
+		return $this->getAdapter()->{$this->generateIfNotExists($functionName)};
+	}
+
+	public function __set($functionName, $mixed)
+	{
+		$this->getAdapter()->{$this->generateIfNotExists($functionName)} = $mixed;
+
+		return $this;
+	}
+
+	public function __isset($functionName)
+	{
+		return $this->functionExists($this->getFqdn($functionName));
+	}
+
+	public function __unset($functionName)
+	{
+		$this->setDefaultBehavior($this->getFqdn($functionName));
+	}
+
+	public function setReflectedFunctionFactory(\closure $factory = null)
+	{
+		$this->reflectedFunctionFactory = $factory ?: function($functionName) { return new \reflectionFunction($functionName); };
+		return $this;
+	}
+
+	public function useClassNamespace($className)
+	{
+		return $this->setDefaultNamespace(substr($className, 0, strrpos($className, '\\')));
+	}
+
+	public function generate($functionName)
+	{
+		$fqdn = $this->getFqdn($functionName);
+		if ($this->functionExists($fqdn) === false)
+		{
+			if (function_exists($fqdn) === true)
+			{
+				throw new exceptions\logic\invalidArgument('Function \'' . $fqdn . '\' already exists');
+			}
+			$lastAntislash = strrpos($fqdn, '\\');
+			$namespace = substr($fqdn, 0, $lastAntislash);
+			$function = substr($fqdn, $lastAntislash + 1);
+			$reflectedFunction = $this->buildReflectedFunction($function);
+			static::defineMockedFunction($namespace, get_class($this), $function, $reflectedFunction);
+		}
+		return $this->setDefaultBehavior($fqdn);
+	}
+
+	public function resetCalls($functionName = null)
+	{
+		static::$adapter->resetCalls($this->getFqdn($functionName));
+		return $this;
+	}
+
+	protected function getFqdn($functionName)
+	{
+		return $this->defaultNamespace . $functionName;
+	}
+
+	protected function generateIfNotExists($functionName)
+	{
+		if (isset($this->{$functionName}) === false)
+		{
+			$this->generate($functionName);
+		}
+		return $this->getFqdn($functionName);
+	}
+
+	protected function setDefaultBehavior($fqdn, \reflectionFunction $reflectedFunction = null)
+	{
+		$function = substr($fqdn, strrpos($fqdn, '\\') + 1);
+		if ($reflectedFunction === null)
+		{
+			$reflectedFunction = $this->buildReflectedFunction($function);
+		}
+		if ($reflectedFunction === null)
+		{
+			$closure = function() { return null; };
+		}
+		else
+		{
+			$closure = eval('return function(' . static::getParametersSignature($reflectedFunction) . ') { return call_user_func_array(\'\\' . $function . '\', ' . static::getParameters($reflectedFunction) . '); };');
+		}
+		static::$adapter->{$fqdn}->setClosure($closure);
+		return $this;
+	}
+
+	protected function functionExists($fqdn)
+	{
+		return (isset(static::$adapter->{$fqdn}) === true);
+	}
+
+	protected static function getParametersSignature(\reflectionFunction $function)
+	{
+		$parameters = array();
+		foreach (self::filterParameters($function) as $parameter)
+		{
+			$parameterCode = self::getParameterType($parameter) . ($parameter->isPassedByReference() == false ? '' : '& ') . '$' . $parameter->getName();
+			switch (true)
+			{
+				case $parameter->isDefaultValueAvailable():
+					$parameterCode .= ' = ' . var_export($parameter->getDefaultValue(), true);
+					break;
+				case $parameter->isOptional():
+					$parameterCode .= ' = null';
+			}
+			$parameters[] = $parameterCode;
+		}
+		return join(', ', $parameters);
+	}
+
+	protected static function getParameters(\reflectionFunction $function)
+	{
+		$parameters = array();
+		foreach (self::filterParameters($function) as $parameter)
+		{
+			$parameters[] = ($parameter->isPassedByReference() === false ? '' : '& ') . '$' . $parameter->getName();
+		}
+		return 'array(' . join(',', $parameters) . ')';
+	}
+
+	protected static function getParameterType(\reflectionParameter $parameter)
+	{
+		switch (true)
+		{
+			case $parameter->isArray():
+				return 'array ';
+			case method_exists($parameter, 'isCallable') && $parameter->isCallable():
+				return 'callable ';
+			case ($class = $parameter->getClass()):
+				return '\\' . $class->getName() . ' ';
+			default:
+				return '';
+		}
+	}
+
+	protected static function defineMockedFunction($namespace, $class, $function, \reflectionFunction $reflectedFunction = null)
+	{
+		eval(sprintf(
+			'namespace %s { function %s(%s) { return \\%s::getAdapter()->invoke(__FUNCTION__, %s); } }',
+			$namespace,
+			$function,
+			$reflectedFunction ? static::getParametersSignature($reflectedFunction) : '',
+			$class,
+			$reflectedFunction ? static::getParameters($reflectedFunction) : 'func_get_args()'
+		));
+	}
+
+	private function buildReflectedFunction($function)
+	{
+		$reflectedFunction = null;
+		try
+		{
+			$reflectedFunction = call_user_func_array($this->reflectedFunctionFactory, array($function));
+		}
+		catch (\exception $exception) {}
+		return $reflectedFunction;
+	}
+
+	private static function filterParameters(\reflectionFunction $function)
+	{
+		return array_filter($function->getParameters(), function($parameter) { return ($parameter->getName() != '...'); });
+	}
+}
--- a/classes/php/mocker/funktion.php
+++ b/classes/php/mocker/funktion.php
@@ -3,6 +3,7 @@
 namespace mageekguy\atoum\php\mocker;
 
 use
+	mageekguy\atoum,
 	mageekguy\atoum\exceptions,
 	mageekguy\atoum\php\mocker
 ;
@@ -40,6 +41,7 @@ class funktion extends mocker
 	public function setReflectedFunctionFactory(\closure $factory = null)
 	{
 		$this->reflectedFunctionFactory = $factory ?: function($functionName) { return new \reflectionFunction($functionName); };
+
 		return $this;
 	}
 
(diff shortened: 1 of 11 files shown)


classes/test.php
```diff
@@ -392,7 +406,8 @@ public function setAssertionManager(test\assertion\manager $assertionManager = n
 			->setHandler('stop', function() use ($test) { if ($test->debugModeIsEnabled() === true) { throw new test\exceptions\stop(); } return $test; })
 			->setHandler('executeOnFailure', function($callback) use ($test) { if ($test->debugModeIsEnabled() === true) { $test->executeOnFailure($callback); } return $test; })
 			->setHandler('dumpOnFailure', function($variable) use ($test) { if ($test->debugModeIsEnabled() === true) { $test->executeOnFailure(function() use ($variable) { var_dump($variable); }); } return $test; })
-			->setPropertyHandler('function', function() use ($test) { return $test->getPhpMocker(); })
+            ->setPropertyHandler('function', function() use ($test) { return $this->getPhpFunctionMocker(); })
```
jubianchi: indentation ;)

classes/test.php
```diff
@@ -392,7 +406,8 @@ public function setAssertionManager(test\assertion\manager $assertionManager = n
...
+            ->setPropertyHandler('constant', function() use ($test) { return $this->getPhpConstantMocker(); })
```
jubianchi: same here

classes/php/mocker/funktion.php
```diff
@@ -0,0 +1,184 @@
...
+
+	public function __unset($functionName)
+	{
+		$this->setDefaultBehavior($this->getFqdn($functionName));
+	}
+
+	public function setReflectedFunctionFactory(\closure $factory = null)
+	{
+		$this->reflectedFunctionFactory = $factory ?: function($functionName) { return new \reflectionFunction($functionName); };
+		return $this;
```
jubianchi: can you add a blank line before the `return` please

classes/php/mocker/funktion.php
```diff
@@ -0,0 +1,184 @@
...
+
+	public function useClassNamespace($className)
+	{
+		return $this->setDefaultNamespace(substr($className, 0, strrpos($className, '\\')));
+	}
+
+	public function generate($functionName)
+	{
+		$fqdn = $this->getFqdn($functionName);
+		if ($this->functionExists($fqdn) === false)
```
jubianchi: blank line before the `if`

classes/php/mocker/funktion.php
```diff
@@ -0,0 +1,184 @@
...
+			{
+				throw new exceptions\logic\invalidArgument('Function \'' . $fqdn . '\' already exists');
+			}
+			$lastAntislash = strrpos($fqdn, '\\');
+			$namespace = substr($fqdn, 0, $lastAntislash);
+			$function = substr($fqdn, $lastAntislash + 1);
+			$reflectedFunction = $this->buildReflectedFunction($function);
+			static::defineMockedFunction($namespace, get_class($this), $function, $reflectedFunction);
+		}
+		return $this->setDefaultBehavior($fqdn);
```
jubianchi: blank line before the `return`

classes/php/mocker/funktion.php
```diff
@@ -0,0 +1,184 @@
...
+	}
+
+	public function resetCalls($functionName = null)
+	{
+		static::$adapter->resetCalls($this->getFqdn($functionName));
+		return $this;
```
jubianchi: blank line before the `return`

classes/php/mocker/funktion.php
```diff
@@ -0,0 +1,184 @@
...
+		return $this->defaultNamespace . $functionName;
+	}
+
+	protected function generateIfNotExists($functionName)
+	{
+		if (isset($this->{$functionName}) === false)
+		{
+			$this->generate($functionName);
+		}
+		return $this->getFqdn($functionName);
```
jubianchi: blank line before the `return`

classes/php/mocker/funktion.php
```diff
@@ -0,0 +1,184 @@
...
+	}
+
+	protected function setDefaultBehavior($fqdn, \reflectionFunction $reflectedFunction = null)
+	{
+		$function = substr($fqdn, strrpos($fqdn, '\\') + 1);
+		if ($reflectedFunction === null)
```
jubianchi: blank line before the `if`

classes/php/mocker/funktion.php
```diff
@@ -0,0 +1,184 @@
...
+		{
+			$reflectedFunction = $this->buildReflectedFunction($function);
+		}
+		if ($reflectedFunction === null)
```
jubianchi: blank line before the `if`

classes/php/mocker/funktion.php
```diff
@@ -0,0 +1,184 @@
...
+		{
+			$closure = function() { return null; };
+		}
+		else
+		{
+			$closure = eval('return function(' . static::getParametersSignature($reflectedFunction) . ') { return call_user_func_array(\'\\' . $function . '\', ' . static::getParameters($reflectedFunction) . '); };');
+		}
+		static::$adapter->{$fqdn}->setClosure($closure);
+		return $this;
```
jubianchi: blank line before the `return`

classes/php/mocker/funktion.php
```diff
@@ -0,0 +1,184 @@
...
+			{
+				case $parameter->isDefaultValueAvailable():
+					$parameterCode .= ' = ' . var_export($parameter->getDefaultValue(), true);
+					break;
+				case $parameter->isOptional():
+					$parameterCode .= ' = null';
+			}
+			$parameters[] = $parameterCode;
+		}
+		return join(', ', $parameters);
```
jubianchi: blank line before the `return`

classes/php/mocker/funktion.php
```diff
@@ -0,0 +1,184 @@
...
+
+	protected function functionExists($fqdn)
+	{
+		return (isset(static::$adapter->{$fqdn}) === true);
+	}
+
+	protected static function getParametersSignature(\reflectionFunction $function)
+	{
+		$parameters = array();
+		foreach (self::filterParameters($function) as $parameter)
```
jubianchi: blank line before the `foreach`

classes/php/mocker/funktion.php
```diff
@@ -0,0 +1,184 @@
...
+			}
+			$parameters[] = $parameterCode;
+		}
+		return join(', ', $parameters);
+	}
+
+	protected static function getParameters(\reflectionFunction $function)
+	{
+		$parameters = array();
+		foreach (self::filterParameters($function) as $parameter)
```
jubianchi: blank line before the `foreach`

classes/php/mocker/funktion.php
```diff
@@ -0,0 +1,184 @@
...
+		{
+			$parameters[] = ($parameter->isPassedByReference() === false ? '' : '& ') . '$' . $parameter->getName();
+		}
+		return 'array(' . join(',', $parameters) . ')';
```
jubianchi: blank line before the `return`

classes/php/mocker/funktion.php
```diff
@@ -0,0 +1,184 @@
...
+
+	private function buildReflectedFunction($function)
+	{
+		$reflectedFunction = null;
+		try
+		{
+			$reflectedFunction = call_user_func_array($this->reflectedFunctionFactory, array($function));
+		}
+		catch (\exception $exception) {}
+		return $reflectedFunction;
```
jubianchi: blank line before the `return`

classes/test.php
Hywan: One morning, you will wake up, and atoum will be PSR-4…

classes/php/mocker/funktion.php
Hywan: Sure!

Hywan:
Tests are failing on PHP5.3 but this not seems to be related to my PR. I will fix them though.

Hywan:
@Hywan No, it's your fault! Should be fixed.

jubianchi:
@Hywan master is stable : https://travis-ci.org/atoum/atoum/branches

Hywan:
Green.

mageekguy:
For the record, i'm not agree to merge this PR in atoum, because this new tool is an open door to bad practice about test creation and code conception.
A mock is an object, even if its behavior can be defined at runtime, so a mock should be as much as possible a black box from the test point of view.
And if you allow the developer to mock a class constant, the mock is not a black box and the developper introduce a tight coupling between its code and the test, which is a bad practice and a problem at middle term.
I'm agree that mock a function introduce the same problem, but mock function was introduced in atoum for a very good reason: PHP is not a full object oriented language, because its native types are not object and it has a procedural API with function and so on.
So, it can be very interesting to mock its functions because at low level, there is no other way to test the code.
However, there is another way than using a constant to know the PHP version (i know, it's just an example, but in my opinion, my point of view is valuable for all use case): just use dependency injection to mock a php version detector!
If you want to do this kind of thing (aka mock a constant), you have an architecture problem in your code and i strongly encourage you to refactor it to separate responsibility between several classes.
One of atom's goal is to help the developper to do thing right, and this tool does not help the developper to do thing right, so i can not be agree for the merge: it's definitely a -1 for me, because atoum should not provide any workaround to avoid refactoring of bad code conception.
My 2 cents.

Hywan:
@mageekguy It is used to mock constants, not class constants.

mageekguy:
So your PR description should be improved.
And you can forget my comment ;).

Hywan:
@mageekguy :heart:

jubianchi:
I'm trying to find a class layout to avoid BC break. What about:

```
abstract class atoum\php\mocker\builder
class atoum\php\mocker\builders\funktion extends atoum\php\mocker\builder
class atoum\php\mocker\builders\constant extends atoum\php\mocker\builder
class_alias($original = atoum\php\mocker\builders\funktion, $alias = atoum\php\mocker)
```

The external API will look identical, thanks to the `class_alias`.

Thoughts @Hywan @mageekguy ?

Hywan:
@jubianchi Not the whole API will be the same. See `atoum\test::setPhpMocker` which become `atoum\test::setPhpFunctionMocker` for instance.

mageekguy:
I don't read your code and i don'have a clear vision of atoum's architecture at this level, but why not using polymorphism here?

Hywan:
And now?

classes/test.php
```diff
@@ -282,14 +284,35 @@ public function getAdapter()
 
 	public function setPhpMocker(php\mocker $phpMocker = null)
 	{
-		$this->phpMocker = $phpMocker ?: new php\mocker();
+		$phpMocker = $phpMocker ?: new php\mocker();
```
Hywan: Not sure it can work here since `php\mocker` does no longer exist.

jubianchi: indeed, it should be php\mocker\funktion

Hywan: And what about:
```php
public function setPhpMocker(php\mocker $phpMocker = null)
{
    return $this->setPhpFunctionMocker($phpMocker);
}
```
I guess the type-hint will warm something…

Hywan:
@jubianchi So, do we need class alias?

classes/test.php
jubianchi: I think the polymorphism, as suggested by @mageekguy, is the best solution. It allows us to gracefully handle both mocker.

jubianchi:
@Hywan I'll try to finish this PR this week, I was busy working on the atom-plugin (which will soon be released)

Hywan:
@jubianchi Excellent.

Hywan:
@jubianchi Do you want me to rebase?

jubianchi:
With the unit tests added, everything should be OK. **But there are some limitations to correctly document.**

Given the following class and test:

```php
<?php

namespace foo {
    class foo {
        public function hello()
        {
            return PHP_VERSION_ID;
        }
    }
}

namespace tests\units\foo {
    use atoum;

    /**
     * @engine inline
     */
    class foo extends atoum
    {
        public function testFoo()
        {
            $this
                ->given($this->newTestedInstance())
                ->then
                    ->variable($this->testedInstance->hello())->isEqualTo(PHP_VERSION_ID)
                ->if($this->constant->PHP_VERSION_ID = uniqid())
                ->then
                    ->variable($this->testedInstance->hello())->isEqualTo(PHP_VERSION_ID)
            ;
        }

        public function testBar()
        {
            $this
                ->given($this->newTestedInstance())
                ->if($this->constant->PHP_VERSION_ID = $mockVersionId = uniqid()) // inline engine will fail here
                ->then
                    ->variable($this->testedInstance->hello())->isEqualTo($mockVersionId)
                ->if($this->constant->PHP_VERSION_ID = $mockVersionId = uniqid()) // isolate/concurrent engines will fail here
                ->then
                    ->variable($this->testedInstance->hello())->isEqualTo($mockVersionId)
            ;
        }
    }
}

```

Note the two comments: depending on the engine, the test will fail (atoum will throw exceptions) at different lines. To "mock" constant, we `define` them in the tested class' namespace which can only be done once.

With the `inline` engine this means we can only stub the constant once per test class. With `isolate` and `concurrent` engines, we can do it once per test method.

jubianchi:
@Hywan yes, please do a rebase to fix merge conflicts :)

Hywan:
@jubianchi I would say we have the same issue with function mocking, isn't it?

Hywan:
@jubianchi Done. Maybe we can squash some commits too.

jubianchi:
@Hywan go ahead ;)

Hywan:
@jubianchi On PHP7 we have a segfault (https://travis-ci.org/atoum/atoum/jobs/92951761), I guess this is an error from Travis PHP build. On PHP5.3, we have a strange error, I can't take a look yet (https://travis-ci.org/atoum/atoum/jobs/92951754). If you have time, feel free to fix it, else, tell me and I will check.

jubianchi:
@Hywan for the segfault I think the only workaround is to disable code coverage for some time until we figure out why it produces segfaults.

Hywan:
@jubianchi I cannot reproduce. This is probably an issue with Travis' build, don't you think?

jubianchi:
@Hywan if you try to reproduce it on OSX I think you won't be able to get the segfault. I never had any problem on OSX but people reported me many segfaults with xdebug on Linux (mainly Debian) and Travis envs. are built on top of it.

Hywan:
@jubianchi Ok. So what's the status of this?

jubianchi:
Seems good to me. One last thing I did not explain in my last comment: there is another limitation on this feature. Take a look at the `testFoo`method in [the previous example](https://github.com/atoum/atoum/pull/520#issuecomment-159293856) : the constant is used once and then stubbed. We then use the constant again and get its original value. 

This is because of how PHP works and resolves names : if we mock something after it has been resolved once, we won't get the mocked value. The same applies to native function mocks.

I'll let you merge here.

Hywan:
@jubianchi Yup, need to update the documentation.

PR #523, opened 2015-11-23
use composer autoloader in atoum if present

agallou:
If we detect that atoum is in a composer environnement,
we load the composer autoloader within the atoum autoloaded.

This will be usefull for extensions :
having to require manually the composer autoloader, in both
bootstrap and .atoum files will not be needed anymore.

We can see this in the extension documentations. 
For example : https://github.com/atoum/ruler-extension

We have to add two lines in the .atoum.php file : 
```php
<?php

// .atoum.php

require_once __DIR__ . DIRECTORY_SEPARATOR . 'vendor' . DIRECTORY_SEPARATOR . 'autoload.php';

$runner->addExtension(new \mageekguy\atoum\ruler\extension($script));
```

### DIFF
--- a/classes/autoloader.php
+++ b/classes/autoloader.php
@@ -263,7 +263,9 @@ class autoloader
 	public static function registerComposerAutoloader()
 	{
 		$composerAutoloaderPath = __DIR__ . '/../../../autoload.php';
-		if (is_file($composerAutoloaderPath)) {
+
+		if (is_file($composerAutoloaderPath))
+		{
 			require_once $composerAutoloaderPath;
 		}
 	}
--- a/classes/autoloader.php
+++ b/classes/autoloader.php
@@ -254,9 +254,18 @@ class autoloader
 		{
 			static::$autoloader = new static();
 			static::$autoloader->register();
+			static::registerComposerAutoloader();
 		}
 
 		return static::$autoloader;
+	}
+
+	public static function registerComposerAutoloader()
+	{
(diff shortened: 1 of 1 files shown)


mikaelrandy:
Seems :+1:

classes/autoloader.php
```diff
@@ -254,11 +254,20 @@ public static function set()
...
+			static::registerComposerAutoloader();
 		}
 
 		return static::$autoloader;
 	}
 
+	public static function registerComposerAutoloader()
+	{
+		$composerAutoloaderPath = __DIR__ . '/../../../autoload.php';
+		if (is_file($composerAutoloaderPath)) {
```
jubianchi: CS: please put the `{` on its own line

jubianchi: please, add a blank line between the variable and the if

jubianchi:
I just removed the target milestone because I think we should take some time to think about this feature: does it produces side-effects in some case?

I'm wondering how (and if) it could do weird things with the symfony bundle for example where we sometimes not directly require composer autoloader but use one from symfony... I don't know... we just have to think twice before merging ;)

BTW thanks @agallou for your patch!

Grummfy:
Perhaps we can have a method to disable this composer autoloading for the integration in sf or other.

Grummfy:
we can have an argument to the command line to the path of the autoloader or something like that?

agallou:
closing in favor of #605 that allows to configure the autoloaded

PR #529, opened 2015-12-07
Allow extension to define configuration

jubianchi:
atoum will use this configuration and pass it to child processes.

@agallou noticed there was something missing around extension configuration while writing the `blackfire-extension`: we can't properly configure the blackfire client as atoum won't forward anything to the child processes actually using the client.

With this patch, extension are free to expose a configuration class and inject an instance of it in atoum when adding the extension so that atoum will forward the configuration to child processes.

For extension developpers, this means they will have to expose a configuration class implementing the provided `mageekguy\atoum\extension\configuration` interface:

```php
<?php

namespace atoum\my\ext
{

    use mageekguy\atoum\observable;
    use mageekguy\atoum\runner;
    use mageekguy\atoum\test;

    class configuration
    {
        private $foo;
        private $bar;

        public function serialize()
        {
            return array(
                'foo' => $this->foo,
                'bar' => $this->bar,
            );
        }

        public function setFoo($v)
        {
            $this->foo = $v;

            return $this;
        }

        public function setBar($v)
        {
            $this->bar = $v;

            return $this;
        }

        public static function unserialize(array $config)
        {
            $configuration = new static;

            $configuration->setFoo($config['foo']);
            $configuration->setBar($config['bar']);

            return $configuration;
        }
    }
}
```

The `configuration` is a bit special: it has to be serialisable as an array representation and should be able to deserialize instance of itself (from the `array` representation).

The extension class will then use this configuration class:

```php
<?php

namespace atoum\my\ext
{

    use mageekguy\atoum\observable;
    use mageekguy\atoum\runner;
    use mageekguy\atoum\test;

    class extension implements \mageekguy\atoum\extension {
        private $configuration;
        private $runner;
        private $test;

        public function __construct()
        {
            $this->configuration = new configuration();
        }

        public function setFoo($v)
        {
            $this->configuration->setFoo($v);

            return $this;
        }

        public function setBar($v)
        {
            $this->configuration->setBar($v);

            return $this;
        }

        public function addToRunner(\mageekguy\atoum\runner $runner)
        {
            $runner->addExtension($this, $this->configuration);

            return $this;
        }

        public function setRunner(runner $runner)
        {
            $this->runner = $runner;

            return $this;
        }

        public function setTest(test $test)
        {
            $this->test = $test;
            $configuration = $test->getExtensionConfiguration($this);

            if ($configuration === null)
            {
                $this->configuration = $configuration;
            }

            return $this;
        }

        public function handleEvent($event, observable $observable)
        {
            return $this;
        }
    }
}
```

When the `setTest` method is called, the extension will be able to fetch its configuration from the `$test` (especially usefull with `concurrent`/`isolate` engines).

For atoum users, this does not change much. Here is an example `.atoum.php` configuration file:

```
<?php

use atoum\my\ext\extension;

$extension = nex extension();

$extension
    ->setFoo('blabla')
    ->setBar('blibli')
    ->addToRunner($runner)
;
```

### DIFF
--- a/classes/extension/configuration.php
+++ b/classes/extension/configuration.php
@@ -1,0 +1,11 @@
+<?php
+
+namespace mageekguy\atoum\extension;
+
+
+interface configuration
+{
+	public function serialize();
+
+	public static function unserialize(array $configuration);
+}
--- a/classes/test/engines/concurrent.php
+++ b/classes/test/engines/concurrent.php
@@ -132,9 +132,35 @@ class concurrent extends test\engine
 				$phpCode .= '$test->getMockGenerator()->disallowUndefinedMethodUsage();';
 			}
 
-			foreach ($test->getExtensions() as $extension)
+			$extensions = $test->getExtensions();
+
+			foreach ($extensions as $extension)
 			{
-				$phpCode .= '$test->addExtension(new ' . get_class($extension) . ');';
+				$configuration = $exception = null;
+				$extensionClass = get_class($extension);
+
+				try
+				{
+					$configuration = $extensions[$extension];
+				}
+				catch (\unexpectedValueException $exception) {}
+
+				if ($exception !== null || $configuration === null)
+				{
+					$phpCode .= '$test->addExtension(new ' . $extensionClass . ');';
+				}
+				else
+				{
+					$configurationClass = get_class($configuration);
+					$serialized = $configuration->serialize();
+
+					if (is_array($serialized) === false)
+					{
+						throw new exceptions\logic('Extension (' . $extensionClass . ') configuration (' . $configurationClass . ') should serialize as an array');
+					}
+
+					$phpCode .= '$test->addExtension(new ' . $extensionClass . ', ' . $configurationClass . '::unserialize(' . var_export($serialized, true) . '));';
+				}
 			}
 
 			$phpCode .=
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 # `dev-master`
 
+* [#529](https://github.com/atoum/atoum/pull/529) Allow extensions to define configuration ([@jubianchi])
 * [#496](https://github.com/atoum/atoum/pull/496) Mock generator supports variadic arguments passed by reference ([@jubianchi])
 * [#496](https://github.com/atoum/atoum/pull/496) Auto generate and inject mocks in test methods ([@jubianchi])
 
--- a/classes/runner.php
+++ b/classes/runner.php
@@ -795,7 +795,7 @@ class runner implements observable
 
 	public function getExtensions()
 	{
-		return iterator_to_array($this->extensions);
+		return $this->extensions;
 	}
 
 	public function removeExtension(atoum\extension $extension)
@@ -818,13 +818,13 @@ class runner implements observable
 	}
 
 
-	public function addExtension(atoum\extension $extension)
+	public function addExtension(atoum\extension $extension, atoum\extension\configuration $configuration = null)
 	{
 		if ($this->extensions->contains($extension) === false)
 		{
 			$extension->setRunner($this);
 
-			$this->extensions->attach($extension);
+			$this->extensions->attach($extension, $configuration);
 
 			$this->addObserver($extension);
 		}
--- a/classes/test.php
+++ b/classes/test.php
@@ -1871,14 +1871,14 @@ abstract class test implements observable, \countable
 
 	public function getExtensions()
 	{
-		return iterator_to_array($this->extensions);
+		return $this->extensions;
 	}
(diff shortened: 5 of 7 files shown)


Hywan:
Do you need a review now?

classes/extension/configuration.php
```diff
@@ -0,0 +1,17 @@
+<?php
+/**
+ * Created by PhpStorm.
+ * User: julien.bianchi
+ * Date: 06/12/2015
+ * Time: 14:19
+ */
+
+namespace mageekguy\atoum\extension;
+
```
Hywan: CS. Too many blank lines.

agallou:
I tested this a have now a runnable test like this for the blackfire-extension : https://gist.github.com/agallou/b9394515b0f865f5ed53

(with this in the .atoum.php) 

```php
$extension = new mageekguy\atoum\blackfire\extension();

$extension
    ->setClientId($_ENV['BLACKFIRE_CLIENT_ID'])
    ->setClientToken($_ENV['BLACKFIRE_CLIENT_TOKEN'])
    ->addToRunner($runner)
;
```

classes/test/engines/concurrent.php
```diff
@@ -132,9 +132,34 @@ public function run(atoum\test $test)
...
+
+			foreach ($extensions as $extension)
 			{
-				$phpCode .= '$test->addExtension(new ' . get_class($extension) . ');';
+				$configuration = $exception = null;
+				$extensionClass = get_class($extension);
+
+				try
+				{
+					$configuration = get_class($extensions[$extension]);
```
agallou: when running an extension without passing the configuration, I get this error : 
```
[.___________________________________________________________][0/1]Fatal error: Call to a member function serialize() on null in /var/www/html/vendor/atoum/atoum/classes/test/engines/concurrent.php on line 154
```

agallou: $extensions[$extension] is null, then get_class returns the current class, and then the test on line 148 is false.

writting  had fixed the problem :
```
$configurationObject = $extensions->getInfo();
                    if (null !== $configurationObject) {
                        $configuration = get_class($configuration);
                    }
```

jubianchi: @agallou thanks! I'll fix that now :)

jubianchi:
@agallou everything should be ok now with or without configuration ;)

Let me know if it works for you and I'll merge this patch so you can make the blackfire extension even more awesome ;)

agallou:
@jubianchi yes, it now works with the previous way of adding and extension, and the new one.

classes/test/engines/concurrent.php
```diff
@@ -132,9 +132,35 @@ public function run(atoum\test $test)
...
+				{
+					$configurationClass = get_class($configuration);
+					$serialized = $configuration->serialize();
+
+					if (is_array($serialized) === false)
+					{
+						throw new exceptions\logic('Extension (' . $extensionClass . ') configuration (' . $configurationClass . ') should serialize as an array');
+					}
+
+					$phpCode .= '$test->addExtension(new ' . $extensionClass . ', ' . $configurationClass . '::unserialize(' . var_export($serialized, true) . '));';
```
agallou: the first tests I made were by passing the clientId and clientToken strings to the configuration. It worked.

but, it may be better to pass the ClientConfiguration object (here, we call an undefined  __set_state method on the ClientConfigration).

what about using serialize/unserialize to do that ? 

By doing this change, passing the ClientConfiguration object to the atoum configuration works : 
```php
$phpCode .= '$test->addExtension(new ' . $extensionClass . ', ' . $configurationClass . '::unserialize(unserialize(' . var_export(serialize($serialized), true) . ')));';
```

@jubianchi what's your thoughts about this ?

jubianchi: let me do the changes ;)

jubianchi: the only drawback of using `(un)serialize` directly is if your configuration instance contains a closure somewhere you will get an error this is why I chose to delegate the serialization to the configuration class: doing so lets the extension developer choose exactly how to serialize the configuration and how to rebuild it from its serialized representation

agallou: ok. i've done the serialize and unserialize calls in the serialize and unserialize methods. This also works; I'll do the PR on blackfire-extension.

agallou: @jubianchi done : https://github.com/agallou/blackfire-extension/pull/3

Hywan:
Is it ready to merge?

jubianchi:
good for me

Hywan:
@jubianchi Can you update the `CHANGELOG.md` file please?

jubianchi:
ping @agallou ;)

Issue #90, opened 2012-06-26
Atoum under Windows


Blackhol3:
In order to complete #84 under Windows, here are some additional tests :

<h3>PHP 5.3.8</h3>

> PHP 5.3.8 (cli) (built: Sep 16 2011 21:28:32)
> Copyright (c) 1997-2011 The PHP Group
> Zend Engine v2.3.0, Copyright (c) 1998-2011 Zend Technologies
> 
> Failure (141 tests, 1026/1026 methods, 1 failure, 0 error, 0 exception) !
> There is 1 failure:
> => mageekguy\atoum\tests\units\mock\stream::testGetSubStream():
> In file phar://XXX/tests/units/classes/mock/stream.php on line 80, mageekguy\atoum\asserters\castToString::match() failed: string(35) 'atoum://4fe9b76d7485d\4fe9b76d74fd2' does not match #^atoum://4fe9b76d7485d\[^\]+$#

<h3>PHP 5.3.6</h3>

> PHP 5.3.6 (cli) (built: Mar 17 2011 10:48:37)
> Copyright (c) 1997-2011 The PHP Group
> Zend Engine v2.3.0, Copyright (c) 1998-2011 Zend Technologies
> 
> Failure (141 tests, 1025/1026 methods, 1 failure, 0 error, 0 exception) !
> There is 1 failure:
> => mageekguy\atoum\tests\units\mock\stream::testGetSubStream():
> In file phar://XXX/tests/units/classes/mock/stream.php on line 80, mageekguy\atoum\asserters\castToString::match() failed: string(35) 'atoum://4fe9c84dba468\4fe9c84dba6f3' does not match #^atoum://4fe9c84dba468\[^\]+$#
> There is 1 uncompleted method:
> => mageekguy\atoum\tests\units\mock\generator::testGenerate() with exit code -1073741819:
> ==> output(0) ""

PR #535, opened 2015-12-15
Asserter: New `resource` asserter!

Hywan:
Example: `$this->resource($variable)->isOfType('stream')`.

### DIFF
--- a/classes/asserters/phpResource.php
+++ b/classes/asserters/phpResource.php
@@ -26,13 +26,15 @@ class phpResource extends asserters\variable
 
 	public function isOfType($type, $failMessage = null)
 	{
-		if (get_resource_type($this->valueIsSet()->value) === $type)
+		$actualType = get_resource_type($this->valueIsSet()->value);
+
+		if ($actualType === $type)
 		{
 			$this->pass();
 		}
 		else
 		{
-			$this->fail($failMessage ?: $this->_('%s is not of type %s', $this, $type));
+			$this->fail(($failMessage ?: $this->_('%s is not of type %s', $this, $type)) . PHP_EOL . $this->diff($actualType));
 		}
 
 		return $this;
--- a/classes/asserters/phpResource.php
+++ b/classes/asserters/phpResource.php
@@ -1,0 +1,40 @@
+<?php
+
+namespace mageekguy\atoum\asserters;
+
+use
+	mageekguy\atoum\asserters
+;
+
+class phpResource extends asserters\variable
+{
+	public function setWith($value)
+	{
+		parent::setWith($value);
+
+		if ($this->analyzer->isResource($this->value) === true)
+		{
+			$this->pass();
+		}
+		else
+		{
+			$this->fail($this->_('%s is not a resource', $this));
+		}
+
+		return $this;
+	}
+
+	public function isOfType($type, $failMessage = null)
+	{
+		if (get_resource_type($this->valueIsSet()->value) === $type)
+		{
+			$this->pass();
+		}
+		else
+		{
+			$this->fail($failMessage ?: $this->_('%s is not of type %s', $this, $type));
+		}
+
+		return $this;
+	}
+}
--- a/classes/asserters/phpResource.php
+++ b/classes/asserters/phpResource.php
@@ -58,7 +58,8 @@ class phpResource extends asserters\variable
 
 	public function __call($name, $arguments)
 	{
-		if ('is' === substr($name, 0, 2)) {
+		if ('is' === substr($name, 0, 2))
+		{
 			$pattern = preg_replace(array('/^is/', '/_/'), array('', '.?'), $name);
 			$pattern = preg_replace_callback(
 				'/([A-Z])([a-z]+)/',
@@ -69,11 +70,12 @@ class phpResource extends asserters\variable
 			);
 			$pattern = '/^' . $pattern . '$/i';
 
-			if (1 === count($arguments)) {
+			if (1 === count($arguments))
+			{
 				return $this->matches($pattern, $arguments[0]);
-			} else {
-				return $this->matches($pattern);
 			}
+
+			return $this->matches($pattern);
 		}
 
 		return parent::__call($name, $arguments);
--- a/classes/asserters/phpResource.php
+++ b/classes/asserters/phpResource.php
@@ -24,6 +24,18 @@ class phpResource extends asserters\variable
 		return $this;
 	}
 
+	public function __get($asserter)
+	{
+		switch (strtolower($asserter))
+		{
+			case 'type':
+				return $this->getTypeAsserter();
+
+			default:
+				return $this->generator->__get($asserter);
+		}
+	}
+
 	public function isOfType($type, $failMessage = null)
 	{
 		$actualType = get_resource_type($this->valueIsSet()->value);
@@ -40,7 +52,7 @@ class phpResource extends asserters\variable
 		return $this;
 	}
 
-	public function matches($pattern, $failMessage = null)
+	protected function matches($pattern, $failMessage = null)
 	{
 		$actualType = get_resource_type($this->valueIsSet()->value);
 
@@ -79,5 +91,10 @@ class phpResource extends asserters\variable
 		}
 
 		return parent::__call($name, $arguments);
+	}
+
+	protected function getTypeAsserter()
+	{
+		return $this->generator->__call('phpString', array(get_resource_type($this->valueIsSet()->value)));
 	}
 }
--- a/classes/asserters/phpResource.php
+++ b/classes/asserters/phpResource.php
@@ -34,7 +34,7 @@ class phpResource extends asserters\variable
 		}
 		else
 		{
-			$this->fail(($failMessage ?: $this->_('%s is not of type %s', $this, $type)) . PHP_EOL . $this->diff($actualType));
+			$this->fail($failMessage ?: $this->_('%s is not of type %s', $this, $type));
 		}
 
 		return $this;
--- a/classes/asserters/phpResource.php
+++ b/classes/asserters/phpResource.php
@@ -24,11 +24,12 @@ class phpResource extends asserters\variable
 		return $this;
 	}
 
-	public function isOfType($type, $failMessage = null)
+	public function matches($type, $failMessage = null)
 	{
+		$pattern = '/^' . $type . '$/i';
 		$actualType = get_resource_type($this->valueIsSet()->value);
 
-		if ($actualType === $type)
+		if (0 !== preg_match($pattern, $actualType))
 		{
 			$this->pass();
 		}
@@ -38,5 +39,27 @@ class phpResource extends asserters\variable
 		}
 
 		return $this;
+	}
+
+	public function __call($name, $arguments)
+	{
+		if ('is' === substr($name, 0, 2)) {
+			$pattern = preg_replace(['/^is/', '/_/'], ['', '.?'], $name);
+			$pattern = preg_replace_callback(
+				'/([A-Z])([a-z]+)/',
+				function ($matches) {
+					return '.?' . strtolower($matches[1]) . $matches[2];
+				},
+				$pattern
+			);
+
+			if (1 === count($arguments)) {
+				return $this->matches($pattern, $arguments[0]);
+			} else {
+				return $this->matches($pattern);
+			}
+		}
+
+		return parent::__call($name, $arguments);
 	}
 }
--- a/classes/asserters/phpResource.php
+++ b/classes/asserters/phpResource.php
@@ -60,7 +60,7 @@ class phpResource extends asserters\variable
 	public function __call($name, $arguments)
 	{
 		if ('is' === substr($name, 0, 2)) {
-			$pattern = preg_replace(['/^is/', '/_/'], ['', '.?'], $name);
+			$pattern = preg_replace(array('/^is/', '/_/'), array('', '.?'), $name);
 			$pattern = preg_replace_callback(
 				'/([A-Z])([a-z]+)/',
 				function ($matches) {
--- a/classes/asserters/phpResource.php
+++ b/classes/asserters/phpResource.php
@@ -40,9 +40,8 @@ class phpResource extends asserters\variable
 		return $this;
 	}
 
-	public function matches($type, $failMessage = null)
+	public function matches($pattern, $failMessage = null)
 	{
-		$pattern = '/^' . $type . '$/i';
 		$actualType = get_resource_type($this->valueIsSet()->value);
 
 		if (0 !== preg_match($pattern, $actualType))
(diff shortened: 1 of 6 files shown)


mikaelrandy:
Huge :+1: on the idea ! I have often thought i need this kind of asserter ;)

And :+1: on the code

classes/asserters/phpResource.php
```diff
@@ -0,0 +1,42 @@
...
+		}
+		else
+		{
+			$this->fail($this->_('%s is not a resource', $this));
+		}
+
+		return $this;
+	}
+
+	public function isOfType($type, $failMessage = null)
```
jubianchi: I would add some helper assertions here like `isStream`, `isMysql`, ... and a wildcard helper via `__get/__call` (`is*`) to allow things like `->isDomXml`, `->isWhatever`, ...

classes/test/adapter/calls.php
```diff
@@ -320,6 +320,10 @@ protected function setCall(adapter\call $call, $position = null)
 			throw new exceptions\logic\invalidArgument('Function is undefined');
 		}
 
+		if (!isset(self::$callsNumber)) {
```
jubianchi: Why ?

This seems a bit unrelated ?

classes/asserters/phpResource.php
Hywan: Maybe we should all native resource, but not the other, like MySQL.

classes/tools/variable/analyzer.php
```diff
@@ -83,6 +83,11 @@ public function isArray($mixed)
 		return (is_array($mixed) === true);
 	}
 
+	public function isResource($mixed)
+	{
+		return (is_resource($mixed) === true);
```
jubianchi: could you also add a `case` in `getTypeOf`please

classes/asserters/phpResource.php
Hywan: http://php.net/resource hmmm the list is very very long :-p.

classes/test/adapter/calls.php
Hywan: It is, or not. If you are testing a resource that of kind `atoum://` stream, you will have an error. See https://github.com/Hywan/atoum/commit/d898c6d6111e98a38555a6ba93b2513198aa3007 for more “explanations”.

classes/tools/variable/analyzer.php
Hywan: Already present.

Hywan: (Go check https://github.com/atoum/atoum/pull/535#discussion-diff-47633946R26).

tests/units/classes/asserters/phpResource.php
```diff
@@ -0,0 +1,97 @@
...
+		$this
+			->given($this->newTestedInstance)
+			->then
+				->object($this->testedInstance->getGenerator())->isEqualTo(new asserter\generator())
+				->object($this->testedInstance->getAnalyzer())->isEqualTo(new variable\analyzer())
+				->object($this->testedInstance->getLocale())->isEqualTo(new atoum\locale())
+				->variable($this->testedInstance->getValue())->isNull()
+				->boolean($this->testedInstance->wasSet())->isFalse()
+
+			->given($this->newTestedInstance($generator = new asserter\generator(), $analyzer = new variable\analyzer(), $locale = new atoum\locale()))
```
jubianchi: why not using the newly introduced auto-mocks ?

```php
public function __construct(asserter\generator $generator, variable\analyzer $analyzer, atoum\locale $locale)
{
    $this
        ->given($this->newTestedInstance($generator, $analyzer, $locale)
        ->then
            ->object($this->testedInstance->getGenerator())-> isIdenticalTo($generator)
            //...
    ;
}
```

tests/units/classes/asserters/phpResource.php
```diff
@@ -0,0 +1,97 @@
...
+				->boolean($this->testedInstance->wasSet())->isFalse()
+		;
+	}
+
+	public function testSetWith()
+	{
+		$this
+			->given(
+				$asserter = $this->newTestedInstance
+					->setAnalyzer($analyzer = new \mock\atoum\tools\variable\analyzer())
```
jubianchi: this mock is not used

tests/units/classes/asserters/phpResource.php
```diff
@@ -0,0 +1,97 @@
...
+					->setLocale($locale = new \mock\atoum\locale()),
+				$this->calling($locale)->_ = $notAResource = uniqid()
+			)
+			->then
+				->exception(function() use ($asserter, & $value) { $asserter->setWith($value = uniqid()); })
+					->isInstanceOf('mageekguy\atoum\asserter\exception')
+					->hasMessage($notAResource)
+				->mock($locale)->call('_')->withArguments('%s is not a resource', $asserter)->once
```
jubianchi: can you indent the `->call` assertion please

tests/units/classes/asserters/phpResource.php
```diff
@@ -0,0 +1,97 @@
...
+				->string($asserter->getValue())->isEqualTo($value)
+
+				->object($asserter->setWith($value = fopen(__FILE__, 'r')))->isIdenticalTo($asserter)
```
jubianchi: did you try using the VFS here ?

classes/asserters/phpResource.php
```diff
@@ -0,0 +1,42 @@
...
+	{
+		$actualType = get_resource_type($this->valueIsSet()->value);
+
+		if ($actualType === $type)
+		{
+			$this->pass();
+		}
+		else
+		{
+			$this->fail(($failMessage ?: $this->_('%s is not of type %s', $this, $type)) . PHP_EOL . $this->diff($actualType));
```
jubianchi: I don't think a diff is useful here for a simple string comparison

tests/units/classes/asserters/phpResource.php
Hywan: Because I just copy/paste test cases from the `integer` class ;-).

Hywan: And your proposal does not work :-).

tests/units/classes/asserters/phpResource.php
Hywan: :+1:

tests/units/classes/asserters/phpResource.php
Hywan: Again, I just copied/pasted from the `integer` test suite ;-). Fixed :+1:

tests/units/classes/asserters/phpResource.php
Hywan: Much heavier than using `fopen(__FILE__)`, and the result is exactly the same :-).

classes/asserters/phpResource.php
Hywan: What do you suggest? Removing the diff?

classes/asserters/phpResource.php
jubianchi: this is why I suggested a wildcard helper ;)

classes/tools/variable/analyzer.php
jubianchi: my bad :)

classes/asserters/phpResource.php
jubianchi: yes

tests/units/classes/asserters/phpResource.php
jubianchi: it works here: https://github.com/atoum/atoum/pull/537/files#diff-f1d31228248836643d3c15038bcb5e73R76

classes/asserters/phpResource.php
Hywan: What to do with space or hyphen? For instance “persistent cubrid connection” is transformed into…?

tests/units/classes/asserters/phpResource.php
jubianchi: ok I forgot a `)` in my snippet but it works like a charm :p

classes/asserters/phpResource.php
Hywan: Same question with “dbx_result_object”.

Hywan: Another question with “gd PS font”.

Hywan: Because we should transform `isPersistentCubridConnection` to `persistent cubric connection`, fine.
But `isDbx_result_object`… crazy.
And `isGdPSFont` will be transformed into `gd ps font`, not `gd PS font`.

:-/

jubianchi: We should compare both values in lowercase. But I agree that we won't be able to cover 100% with helper like `isMysql`or `isStream` or the wildcard helper. This is why we still need the `isOfType` asssertion.

classes/test/adapter/calls.php
```diff
@@ -320,6 +320,12 @@ protected function setCall(adapter\call $call, $position = null)
 			throw new exceptions\logic\invalidArgument('Function is undefined');
 		}
 
+		// A bug in PHP removes this static attribute when destructing all the
+		// object. So accessing it create an error. This is a workaround.
+		if (!isset(self::$callsNumber)) {
```
jubianchi: https://github.com/Hywan/atoum/pull/4

tests/units/classes/asserters/phpResource.php
Hywan: @jubianchi Can you make a PR, I don't understand your purpose (and I wasn't speaking about the syntax error ;-)).

classes/asserters/phpResource.php
Hywan: Fixed.

Hywan: I have found a trick to support all the possible kind of resource types.

classes/test/adapter/calls.php
Hywan: It seems to break the tests :-/.

tests/units/classes/asserters/phpResource.php
```diff
@@ -0,0 +1,138 @@
...
+						->withArguments('%s is not a resource', $asserter)
+						->once
+				->string($asserter->getValue())->isEqualTo($value)
+
+				->object($asserter->setWith($value = fopen(atoum\mock\streams\fs\file::get(), 'r')))->isIdenticalTo($asserter)
+				->resource($asserter->getValue())->isEqualTo($value)
+		;
+	}
+
+	public function testIsOfType()
```
jubianchi: this test should be renamed to `testMatches`

jubianchi:
@Hywan with your last commits, failure message are a bit weird. For example: `Resource id #90 of type stream is not of type .?foo.?bar`

I think it's better to not rely on pattern matching at all and only handle common cases (eg split camel case and transform to space-separated string).

Edge-cases would be handled by the user using is `isOfType` assertion.

What do you think ?

tests/units/classes/asserters/phpResource.php
Hywan: Done.

Hywan:
@jubianchi Here is what I did: `isOfType` with a regular string comparison, and `matches` with regular expression, and `is*` asserters fallback to `matches`. This is pretty logical this way.

Hywan:
Everything seems good. Waiting on tests.

Hywan:
Hmm, still a segfault…

Hywan:
@jubianchi The segfault was caused by using VFS instead of a plain `fopen(__FILE__, 'r')` ;-).

Hywan:
Ready to merge! Here is final API:
```php
$this
    ->resource($fp)
        ->isOfType('stream')
        ->isIdenticalTo($anotherFp);
```
(Because we inherit from `variable` we can use `isIdenticalTo` & co.).
```php
$this
    ->resource($fp)
        ->isOfType('stream')
        ->matches('stream')
        ->isStream()
```
All these examples are similar.
However, is `resource::matches` has the same behavior than `string::matches`? No. `string::matches` expect a full PCRE expression, like `/stream/` for instance, while here it is just `stream`. This is not consistent. I will update that.

Hywan:
The final API:
Ready to merge! Here is the current API:
```php
$this
    ->resource($fp)
        ->isOfType('stream')
        ->isIdenticalTo($anotherFp)
        ->matches('/^stream$/')
        ->isStream()
```

Hywan:
Readyyy :-].

jubianchi:
I think the API is a bit misleading, especially, the `matches` assertion does not seem appropriate to me (not clear enough).

User might expect that atoum will try to match the pattern against the stream contents.

To assert on stream type, what do you think of implementing something like:

```php
<?php

$this
    ->resource($fp)
        ->isOfType('stream')
        ->isStream()
        ->type
            ->matches('/^stream$/')
```

All assertions here are equivalent, they will assert on the stream type. The only difference with your proposal is the `->type->...` form: we fallback to a `string` asserter to assert on the type value letting us do more assertions (`isEqualTo`, `matches`, `startsWith`, ...)

Hywan:
@jubianchi How to implement this without breaking the “assertion flow”? For instance `->resource(…)->type->isEqualTo(…)->array(…)->…`?

jubianchi:
@Hywan see:

* https://github.com/atoum/atoum/blob/master/classes/asserters/exception.php#L22-L23
* https://github.com/atoum/atoum/blob/master/classes/asserters/exception.php#L149-L152

;)

classes/asserters/phpResource.php
```diff
@@ -0,0 +1,81 @@
...
+		{
+			$this->fail($failMessage ?: $this->_('%s does not match %s', $this, $pattern));
+		}
+
+		return $this;
+	}
+
+	public function __call($name, $arguments)
+	{
+		if ('is' === substr($name, 0, 2)) {
```
jubianchi: `{` should be on its own line

classes/asserters/phpResource.php
```diff
@@ -0,0 +1,81 @@
...
+			$pattern = preg_replace_callback(
+				'/([A-Z])([a-z]+)/',
+				function ($matches) {
+					return '.?' . strtolower($matches[1]) . $matches[2];
+				},
+				$pattern
+			);
+			$pattern = '/^' . $pattern . '$/i';
+
+			if (1 === count($arguments)) {
```
jubianchi: `{` should be on its own line

classes/asserters/phpResource.php
```diff
@@ -0,0 +1,81 @@
...
+				return $this->matches($pattern, $arguments[0]);
+			} else {
```
jubianchi: `}` and `{` should be on their own line

jubianchi:
@Hywan do you think you can make your PR ready before the 2.5.0 due date (01-07-2016) ?

If no, do you allow me to apply the required refactoring to your branch and merge it ?

Hywan:
@jubianchi I will try for the next 30mn ;-).

Hywan:
@jubianchi Normally it's done and it works. The test fails because “the branch does not exist”, only for some version of PHP. I guess this is an infrastructure issue, and not related to atoum or this PR. The last commit was necessary and it will run a new fresh test campaign. Wait and see.

Hywan:
Green :white_check_mark: \o/.

jubianchi:
merge in 601663a72fafe9b

PR #227, opened 2013-06-25
atoum-phpunit bridge

jubianchi:
This PR is a work in progress. It aims to implement a bridge allowing atoum to run PHPUnit tests.
Currently, this supports the following from PHPUnit:
## Assertions

<table>
<tr>
<td>assertFalse</td><td>assertTrue</td><td>assertNull</td><td>assertNotNull</td>
</tr>
<tr>
<td>assertEquals</td><td>assertNotEquals</td><td>assertContains</td><td>assertNotContains</td>
</tr>
<tr>
<td>assertSame</td><td>assertInstanceOf</td><td>assertNotInstanceof</td><td>assertArrayHasKey</td>
</tr>
<tr>
<td>assertCount</td><td>assertFileExists</td><td>assertFileNotExists</td><td>assertStringStartsWith</td>
</tr>
<tr>
<td>assertAttributeEquals</td>
</tr>
</table>

## Mock

<table>
<tr>
<td>getMock</td><td>getMockForAbstractClass</td><td>getMockBuilder</td>
</tr>
</table>

### Expectations

<table>
<tr>
<td>expects</td><td>any</td><td>exactly</td><td>once</td><td>atLeastOnce</td><td>at</td><td>never</td>
</tr>
<tr>
<td>with</td><td>equalTo</td><td>identicalTo</td>
</tr>
</table>

### Definition

<table>
<tr>
<td>will</td><td>returnValue</td><td>returnVallback</td><td>onConsecutiveCalls</td><td>throwException</td>
</tr>
</table>

So basically, something like this will work with atoum:

``` php
<?php
$mock = $this->getMock('\\Foo\\Bar');
$mock
    ->expect($this->any())
    ->method('baz')
    ->with('arg1', 'arg2')
    ->will($this->returnValue('returnval'))
;

//...
```
### Misc.

<table>
<tr>
<td>@group</td><td>markTestSkipped</td><td>fail</td>
</tr>
</table>

## How to use:

I ported two test suites as examples:
- Composer
  - jubianchi/composer: https://github.com/jubianchi/composer/compare/composer:master...master
  - travis builds: https://travis-ci.org/jubianchi/composer
- Twig
  - jubianchi/Twig: https://github.com/jubianchi/Twig/compare/fabpot:master...master
  - travis builds: https://travis-ci.org/jubianchi/Twig

Enabling atoum on a PHPUnit test suite is quite simple:
[...]

PR #539, opened 2015-12-21
[WiP] $this->newMockInstance($class)

Grummfy:
Hello,
following the discussion on IRC, here is the start of the PR
I think we should speak about it, and try with our existing base of code to be sure nothing is broken.

the purpose is to get a mock instance without using new \mock\.... 
Why, mainly to make IDE happier, also because people prefer to have another way to make mock instance.

How it will work:
Normal way : new \mock\a\b\c();
New ways :
* $this->mockGenerator->getNewMockInstance('\\a\\b\\c');
* $this->newMockInstance('\\a\\b\\c');

### DIFF
--- a/classes/mock/generator.php
+++ b/classes/mock/generator.php
@@ -201,6 +201,26 @@ class generator
 
 		return $this;
 	}
+	
+	public function getNewMockInstance($class, $mockNamespace = null, $mockClass = null, array $constructorArguments = array())
+	{
+		$this->generate($class, $mockNamespace, $mockClass);
+		
+		if ($mockNamespace === null)
+		{
+			$mockNamespace = $this->getNamespace($class);
+		}
+
+		$class = '\\' . ltrim($class, '\\');
+
+		if ($mockClass === null)
+		{
+			$mockClass = self::getClassName($class);
+		}
+		
+		$reflectionClass = call_user_func($this->reflectionClassFactory, $mockNamespace . '\\' . $mockClass);
+		return $reflectionClass->newInstanceArgs($constructorArguments);
+	}
 
 	public function methodIsMockable(\reflectionMethod $method)
 	{
--- a/classes/mock/generator.php
+++ b/classes/mock/generator.php
@@ -201,26 +201,6 @@ class generator
 
 		return $this;
 	}
-	
-	public function getNewMockInstance($class, $mockNamespace = null, $mockClass = null, array $constructorArguments = array())
-	{
-		$this->generate($class, $mockNamespace, $mockClass);
-		
-		if ($mockNamespace === null)
-		{
-			$mockNamespace = $this->getNamespace($class);
-		}
-
-		$class = '\\' . ltrim($class, '\\');
-
-		if ($mockClass === null)
-		{
-			$mockClass = self::getClassName($class);
-		}
-		
-		$reflectionClass = call_user_func($this->reflectionClassFactory, $mockNamespace . '\\' . $mockClass);
-		return $reflectionClass->newInstanceArgs($constructorArguments);
-	}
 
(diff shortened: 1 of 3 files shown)


Grummfy:
before gooing further I would like to have some return from @Hywan , @jubianchi  or anyone else
thanks

Grummfy:
Note, the PR https://github.com/atoum/atoum/pull/227 already have the given behaviour!

classes/mock/generator.php
```diff
@@ -201,6 +201,26 @@ public function generate($class, $mockNamespace = null, $mockClass = null)
 
 		return $this;
 	}
+	
+	public function getNewMockInstance($class, $mockNamespace = null, $mockClass = null, array $constructorArguments = array())
```
jubianchi: I would remove this method as the purpose of the mock generator is to generate mock classes, not instances.

classes/test.php
```diff
@@ -1242,6 +1242,11 @@ public function runTestMethod($testMethod, array $tags = array())
 						}
 					);
 
+					$this->assertionManager->setPropertyHandler('newMockInstance', function($class, $mockNamespace = null, $mockClass = null, array $constructorArguments = array()) {
+							return $this->mockGenerator->getNewMockInstance($class, $mockNamespace, $mockClass, $constructorArguments);
```
jubianchi: Here i would use a `new`like that: 

```php
$namespace = $this->mockGenerator->getDefaultNamespace();
$classname = rtrim($namespace, '\\') . '\\' . ltrim($class, '\\');

return new $classname;

```

This will cover most use-cases. If the user provides some constructor arguments we might rely on reflection to instanciate the mock but this should not be the default behavior.

Grummfy:
@jubianchi for me the phpunit bridge could be used (because there is something like this in it) instead of this, no?

jubianchi:
@Grummfy If I did the same implementation in the bridge (honestly I don't remember as it's been a while since I checked out the branch) I think it was a mistake :)

Grummfy:
No you didn't make the same implementation but something like this.
We could probably use each other to make something usable in both side

jubianchi:
@Grummfy I'm extracting the bridge as an extension and it will benefit from the feature you implemented ;)

Grummfy:
OKI doki (Yihaaaa).
I will try to refactor it with your previous comments.

classes/mock/generator.php
Grummfy: So, I should create a method that will return "$mockNamespace . '\\' . $mockClass" to let the generator generate the class and create the instance in another place.

jubianchi:
@Grummfy do you think you can make your PR ready before the 2.5.0 due date (01-07-2016) ?

If no, do you allow me to apply the required refactoring to your branch and merge it ?

Grummfy:
"$this->mockGenerator->getNewMockInstance('\a\b\c');" a été supprimé par contr ej'ai du mal a voir comment écrire n unit test au sujet de cette méthode

Grummfy:
It should be finish now but still don't know how to test it (unit test it)

Test demo (totally useless in this context) : https://gist.github.com/Grummfy/04b2f9209583abfcd289

Grummfy:
@jubianchi if it need refactoring feel free to do it, but please say to me what was wrong to improve it for the next time

jubianchi:
closing this one in favor of #550

Grummfy:
Thanks.
Le 6 janv. 2016 10:09 PM, "Julien BIANCHI" <notifications@github.com> a
écrit :

jubianchi:
@Grummfy thank you ;) I made some comment on the code I added to let you know exactly what I did

PR #541, opened 2016-01-01
New asserters

jubianchi:
This PR adds two new asserters:

* `castToArray` to cast any value to an array and assert on it:

```php
$this->castToArray($actual)->string[0]->isEqualTo($expected);

// OR

$this->string($actual)->toArray->string[0]->isEqualTo($expected);
```

* `iterator` to assert on any instance from a class implemeting the `iterator` interface:

```php
$this->iterator($object)->hasSize(5)
```

Those two asserters can also work together to ease assertions on `iterator` content:

```php
$this
    ->iterator($actual)
    ->hasSize(5)
    ->toArray
        ->object[0]->isEqualTo($expected)
        // ...
```

_Those two asserters have been extracted from #227_

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 # `dev-master`
 
+* [#541](https://github.com/atoum/atoum/pull/541) New `toArray` method on the `phpString` and `object` asserters to cast value to an array ([@jubianchi])
 * [#541](https://github.com/atoum/atoum/pull/541) New `iterator` asserter ([@jubianchi])
 * [#541](https://github.com/atoum/atoum/pull/541) New `castToArray` asserter ([@jubianchi])
 * [#535](https://github.com/atoum/atoum/pull/535) New `resource` asserter group (with `isOfType` or `is*` wildcard like `isStream`) ([@hywan])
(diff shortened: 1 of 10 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 # `dev-master`
 
+* [#541](https://github.com/atoum/atoum/pull/541) New `toArray` method on the `phpString` and `object` asserters to cast value to an array ([@jubianchi])
+* [#541](https://github.com/atoum/atoum/pull/541) New `iterator` asserter ([@jubianchi])
+* [#541](https://github.com/atoum/atoum/pull/541) New `castToArray` asserter ([@jubianchi])
```
mikaelrandy: I think it the PR description should be self-sufficient, and the changelog should only contain an unique link to the PR

mikaelrandy:
If i understand the code, the `toArray()` asserter only work for string and `iterator()` asserter ?

jubianchi:
@mikaelrandy nope, it will cast any value to an array: https://github.com/atoum/atoum/pull/541/files#diff-edac2b3982c4cd87b1fbe7e2dc15a9e3R54

mikaelrandy:
@jubianchi oh ! right.

Should :+1: to me

PR #550, opened 2016-01-06
Get mock instance

jubianchi:


classes/test.php
```diff
@@ -411,6 +411,27 @@ public function setAssertionManager(test\assertion\manager $assertionManager = n
 			->setHandler('mockGenerator', function() use ($test) { return $test->getMockGenerator(); })
 			->setHandler('mockClass', function($class, $mockNamespace = null, $mockClass = null) use ($test) { $test->getMockGenerator()->generate($class, $mockNamespace, $mockClass); return $test; })
 			->setHandler('mockTestedClass', function($mockNamespace = null, $mockClass = null) use ($test) { $test->getMockGenerator()->generate($test->getTestedClassName(), $mockNamespace, $mockClass); return $test; })
+			->setHandler('newMockInstance', function($class, $mockNamespace = null, $mockClass = null, array $constructorArguments = null) use ($test)
```
jubianchi: @Grummfy I applied some refactoring here...

classes/test.php
```diff
@@ -411,6 +411,27 @@ public function setAssertionManager(test\assertion\manager $assertionManager = n
...
+				{
+					$mockNamespace = trim($mockNamespace ?: $test->getMockGenerator()->getDefaultNamespace(), '\\');
+					$mockClass = trim($mockClass ?: $class, '\\');
+					$className = $mockNamespace . '\\' . $mockClass;
+
+					if (class_exists($className) === false)
```
jubianchi: we have to check if the class exists before generating it as the mock generator does not check that for us

classes/test.php
```diff
@@ -411,6 +411,27 @@ public function setAssertionManager(test\assertion\manager $assertionManager = n
...
+					{
+						$test->mockClass($class, $mockNamespace, $mockClass);
```
jubianchi: here we use a previously declared handler to generate the mock class

classes/test.php
```diff
@@ -411,6 +411,27 @@ public function setAssertionManager(test\assertion\manager $assertionManager = n
...
+					}
+
+					if ($constructorArguments !== null)
```
jubianchi: we only rely on reflection if needed

classes/test.php
```diff
@@ -411,6 +411,27 @@ public function setAssertionManager(test\assertion\manager $assertionManager = n
...
+					{
+						$reflectionClass = new \reflectionClass($className);
+
+						return $reflectionClass->newInstanceArgs($constructorArguments);
+					}
+
+					return new $className();
+				}
```
jubianchi: @Grummfy no big change here, your code was already correct ;)

tests/units/classes/test.php
```diff
@@ -1245,5 +1250,51 @@ public function testStaticMethod()
 						->isEqualTo(array($return1, $return2, $return3))
 			;
 		}
+
+		public function testNewMockInstance()
+		{
+			$this
+				->if($test = new emptyTest())
+				->then
+					->object($mock = $test->newMockInstance('stdClass'))
```
jubianchi: Here we test all the possible call to `newMockInstance` and check that the mock is correct

tests/units/classes/test.php
```diff
@@ -1245,5 +1250,51 @@ public function testStaticMethod()
...
+						->isInstanceOf('stdClass')
+						->isNotIdenticalTo($mock)
+					->object($test->newMockInstance('stdClass', 'foobar'))
+						->isInstanceOf('foobar\stdClass')
+						->isInstanceOf('stdClass')
+					->object($test->newMockInstance('stdClass', 'foo', 'bar'))
+						->isInstanceOf('foo\bar')
+						->isInstanceOf('stdClass')
+
+				->given($arguments = array($firstArgument = uniqid(), $secondArgument = rand(0, PHP_INT_MAX)))
```
jubianchi: Here we check that when we pass constructor arguments they are correctly passed to the mock constructor

tests/units/classes/test.php
```diff
@@ -1245,5 +1250,51 @@ public function testStaticMethod()
...
+				->then
+					->object($mock = $test->newMockInstance('mageekguy\atoum\dummy', null, null, $arguments))
+						->isInstanceOf('mock\mageekguy\atoum\dummy')
+						->isInstanceOf('mageekguy\atoum\dummy')
+					->mock($mock)
+						->call('__construct')->withArguments($firstArgument, $secondArgument)->once
+
+				->given($arguments = array(uniqid(), rand(0, PHP_INT_MAX), $controller = new mock\controller()))
```
jubianchi: We add an additional argument which is a mock controller and check that atoum correctly injected it into the mock instance (via constructor arguments)

tests/units/classes/test.php
```diff
@@ -1245,5 +1250,51 @@ public function testStaticMethod()
...
+				->then
+					->object($mock = $test->newMockInstance('mageekguy\atoum\dummy', null, null, $arguments))
+						->isInstanceOf('mock\mageekguy\atoum\dummy')
+						->isInstanceOf('mageekguy\atoum\dummy')
+					->object($mock->getMockController())->isIdenticalTo($controller)
+
+				->given(
+					$arguments = array(uniqid(), rand(0, PHP_INT_MAX)),
+					$controller = new mock\controller()
```
jubianchi: We add an additional argument which is a mock controller and check that atoum correctly injected it into the mock instance (via automagic injection)

tests/units/classes/test.php
```diff
@@ -1245,5 +1250,51 @@ public function testStaticMethod()
 						->isEqualTo(array($return1, $return2, $return3))
 			;
 		}
+
+		public function testNewMockInstance()
+		{
+			$this
+				->if($test = new emptyTest())
```
jubianchi: we use a dummy `emptyTest` class (declared at the top of this file) as the tested class

classes/test.php
Grummfy: I have seen that in mockgenerator we can defined another reflexion class. Perhaps we should inject it here too?

jubianchi: yes at some places we use a reflectionClassFactory (for example in the constructor of this class) to deeply test code behavior but it was not required here: those tests are tightly coupled to implementation, I prefered to only check external behaviors

PR #551, opened 2016-01-08
Disable xdebug on php7 when testing the phar

jubianchi:
This will fix the phar build on travis with PHP7.

Ideally, this PR should go in a 2.5.1 but it not a blocker ATM we can postpone it until 2.6.0


Hywan:
:+1:

PR #557, opened 2016-01-18
Make the autoloader handle traits

jubianchi:


jubianchi:
Here is the PR for the `1.x``

@atoum/rms :warning: the tag has to be done from the `1.x` branch

PR #567, opened 2016-01-31
Extract loop logic from runner and add a looper interface

jubianchi:
See #555 for the original idea and PR by @agallou.

I just extracted the loop logic (asking the user to hit enter) and made an interface allowing new strategy to be plugged. Here is another one I implemented to test the feature:

```php
<?php

namespace mageekguy\atoum\scripts\runner\loopers;

use mageekguy\atoum\scripts\runner\looper;

class signal implements looper
{
	private $runAgain = false;

	public function __construct()
	{
		$runAgain = & $this->runAgain;

		pcntl_signal(SIGUSR1, function() use  (& $runAgain) { $runAgain = true; });
	}

	public function runAgain()
	{
		while ($this->runAgain === false)
		{
			usleep(100);

			pcntl_signal_dispatch();
		}

		$this->runAgain = false;

		return true;
	}
}

```

### DIFF
--- a/classes/scripts/runner.php
+++ b/classes/scripts/runner.php
@@ -27,11 +27,12 @@ class runner extends atoum\script\configurable
 	protected $tags = array();
 	protected $methods = array();
 	protected $loop = false;
+	protected $looper;
 
 	protected static $autorunner = true;
 	protected static $runnerFile = null;
 
-	public function __construct($name, atoum\adapter $adapter = null)
+	public function __construct($name, atoum\adapter $adapter = null, atoum\scripts\runner\looper $looper = null)
 	{
 		parent::__construct($name, $adapter);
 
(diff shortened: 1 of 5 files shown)


classes/scripts/runner.php
```diff
@@ -132,6 +134,18 @@ public function getDefaultReportFactory()
 		return $this->defaultReportFactory;
 	}
 
+	public function setLooper(atoum\scripts\runner\looper $looper = null)
+	{
+		$this->looper = $looper ?: new atoum\scripts\runner\loopers\signal($this->prompt, $this->outputWriter, $this->cli, $this->locale);
```
agallou: ce n'est pas plutôt atoum\scripts\runner\loopers\prompt ?

jubianchi: bien vu !

PR #569, opened 2016-02-02
Cache resolved asserters

jubianchi:
This tiny patch let us improve performances on atoum tests (this is really clear with inline engine).

_All the metrics above were generated using the [phpunit-extension](https://github.com/atoum/phpunit-extension) and the [PHP-DI test suite](https://github.com/jubianchi/PHP-DI/compare/e6f30a337cdbff13a1dd9ac73c59654ce44177ab...HEAD) (`vendor/bin/atoum -d tests/UnitTest/ tests/IntegrationTest`)_

### Before

```
PHP 7.0.0 (cli) (built: Dec  2 2015 13:06:23) ( NTS )
Copyright (c) 1997-2015 The PHP Group
Zend Engine v3.0.0, Copyright (c) 1998-2015 Zend Technologies
    with blackfire v1.9.0, https://blackfire.io, by Blackfireio Inc.

> Total tests duration: 29.46 seconds.
> Total tests memory usage: 4.00 Mb.
> Running duration: 29.53 seconds.
```

### After

```
PHP 7.0.0 (cli) (built: Dec  2 2015 13:06:23) ( NTS )
Copyright (c) 1997-2015 The PHP Group
Zend Engine v3.0.0, Copyright (c) 1998-2015 Zend Technologies
    with blackfire v1.9.0, https://blackfire.io, by Blackfireio Inc.

> Total tests duration: 1.84 seconds.
> Total tests memory usage: 4.00 Mb.
> Running duration: 1.89 seconds.
```

Here is the diff generated with blackfire (`blackfire run --samples=5 vendor/bin/atoum -d tests/UnitTest/ tests/IntegrationTest`):

blackfire

### DIFF
--- a/classes/asserter/resolver.php
+++ b/classes/asserter/resolver.php
@@ -15,6 +15,7 @@ class resolver
 	protected $baseClass = '';
 	protected $namespaces = array();
 	private $analyzer;
+	private $resolved = array();
 
 	public function __construct($baseClass = null, $namespace = null, analyzer $analyzer = null)
 	{
@@ -63,6 +64,10 @@ class resolver
 
 	public function resolve($asserter)
 	{
+		if (isset($this->resolved[$asserter])) {
+			return $this->resolved[$asserter];
+		}
+
 		if (false === $this->analyzer->isValidNamespace($asserter))
 		{
 			return null;
@@ -90,6 +95,8 @@ class resolver
 				break;
 			}
 		}
+
+		$this->resolved[$asserter] = $class;
 
 		return $class;
 	}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 # `dev-master`
 
+* [#569](https://github.com/atoum/atoum/pull/569) Use in-memory cache for resolved asserters ([@jubianchi])
 * [#567](https://github.com/atoum/atoum/pull/567) Extract loop logic from runner and add a looper interface ([@jubianchi], [@agallou])
 
 # 2.5.2 - 2016-01-28


Hywan:
:+1:
Impressive how it boosts performance :-/.

Hywan:
Good job!

mikaelrandy:
:+1:

Mandragora2:
:+1: :heart:

dupuchba:
:+1:

classes/asserter/resolver.php
```diff
@@ -63,6 +64,10 @@ public function getNamespaces()
 
 	public function resolve($asserter)
```
mageekguy: Alternative version:

```
	public function resolve($asserter)
	{
		if (! isset($this->resolved[$asserter]))
		{
			$class = null;

			if ($this->analyzer->isValidNamespace($asserter))
			{
				if (strpos($asserter, '\\') !== false)
				{
					$class = $this->checkClass($asserter);
				}
				else foreach ($this->namespaces as $namespace)
				{
					switch (true)
					{
						case $class = $this->checkClass($namespace . '\\' . $asserter):
						case $class = $this->checkClass($namespace . '\\php' . ucfirst($asserter));
							break 2;
					}
				}
			}

			$this->resolved[$asserter] = $class;
		}

		return $this->resolved[$asserter];
	}
```

classes/asserter/resolver.php
```diff
@@ -15,6 +15,7 @@ class resolver
 	protected $baseClass = '';
 	protected $namespaces = array();
 	private $analyzer;
+	private $resolved = array();
```
mageekguy: Maybe a static attribute is better here, to shared the cache between all instances of the resolver (even if there is only one instance in atoum, in theory).

mageekguy:
Very good job, the result is really impressive!

SylvG:
:+1:

mageekguy:
And again, a huge warning: the impact of this patch will be less impressive with other engines (fork is main bottleneck with other engines, and with a fork, the cache is not so efficient because it will be recreated in each subprocess)!

HitoTech:
:+1:

classes/asserter/resolver.php
jubianchi: right!

jubianchi:
@mageekguy totally right!

I'm already working on other performance improvements targeted at concurrent and isolate engines.

Hywan:
@mageekguy, @jubianchi Maybe we can share the cache between process but it can really slow down the execution due to concurrent I/O. But inside the same test case (so the same process), it will speed up anyway as soon as we use an assertion more than once.

mageekguy:
@Hywan Cost of serialization/deserialization to transfert cache between each processus will cancel the cache benefit.

Hywan:
@mageekguy Except if the cache is in a shared memory page: http://php.net/shmop.

mageekguy:
@Hywan Shared memory page imply specific PHP extension which is rarely available "out of the box" on a workstation and it seems to me that it is unavailable on Windows. Moreover, you must manage lock to use it to avoid race condition. So, this solution add lot of complexity in the code in regards of a small impact on performance and a lack of portability. Not sure that wasting time to implements this kind of tradeoff is a good move, or only for the fun.

Hywan:
@mageekguy It works on Windows :-).

PR #570, opened 2016-02-06
Add a confirmation before generating tag

### LINKED ISSUE #563: The pusher script should ask a confirmation about the version it will create
The [pusher script](https://github.com/atoum/atoum/blob/master/scripts/git/pusher.php) do not ask confirmation (or propose a --dry-run) mode to check it correct usage.

mikaelrandy:
Close #563 

To avoid mistake while using git/pusher script, add a confirmation who display tag to be generated before doing tag generation.

Add a "--force" mode to avoid any confirmation and allow to use this script the same way before this modification

### DIFF
--- a/classes/scripts/git/pusher.php
+++ b/classes/scripts/git/pusher.php
@@ -25,6 +25,7 @@ class pusher extends script\configurable
 	protected $workingDirectory = '';
 	protected $taggerEngine = null;
 	protected $git = null;
+	protected $forceMode = false;
 	protected $tagMajorVersion = false;
 	protected $tagMinorVersion = false;
 
@@ -110,6 +111,18 @@ class pusher extends script\configurable
 		return $this->git;
 	}
 
+	public function setForceMode($force=true)
+	{
+		$this->forceMode = $force;
+
+		return $this;
+	}
+
+	public function getForceMode()
+	{
+		return $this->forceMode;
+	}
+
 	public function tagMajorVersion()
 	{
 		$this->tagMajorVersion = true;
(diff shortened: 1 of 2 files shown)


classes/scripts/git/pusher.php
```diff
@@ -200,6 +200,10 @@ protected function doRun()
 
 			$tag = $this->getNextVersion(trim($tag));
 
+			if($this->prompt(sprintf($this->locale->_("Version to generate : '%s'. Type 'Y' to confirm."), $tag)) !== 'Y') {
```
GuillaumeDievart: The brace must be on a new line, with a space between `if` and `(`

mikaelrandy: Done, thanks

classes/scripts/git/pusher.php
```diff
@@ -200,6 +220,11 @@ protected function doRun()
 
 			$tag = $this->getNextVersion(trim($tag));
 
+			if ($this->getForceMode() === false && $this->prompt(sprintf($this->locale->_("Version to generate : '%s'. Type 'Y' to confirm."), $tag)) !== 'Y')
```
jubianchi: What about:

> You are about to push the '%s' version. Type 'Y' to confirm."

mikaelrandy: :+1:

jubianchi:
@mikaelrandy should we add a changelog entry for this ?

mikaelrandy:
@jubianchi i dont think it's necessary.
This code is not usefull for using atoum, but for releasing atoum.

mikaelrandy:
:+1:

PR #576, opened 2016-02-20
Issue mkdir

remicollet:
I encounter a strange error:

     Error: copy(/work/GIT/galette/tests/coverage/screen.css): failed to open stream: Aucun fichier ou dossier de ce type in /usr/share/atoum/classes/adapter.php at line 18

After digging a little, I understand that the "coverage" destination directory doesn't exists
This seems quite common (it exists on the dev working clone, but also in .gitignore, so missing on any fresh clone).

Could be a nice idea to create it during the run

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,7 +61,7 @@ $this
 	->then
 		->boolean($testedInstance->doSomething())->isTrue()
 		->mock($aMock)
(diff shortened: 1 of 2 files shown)


remicollet:
grr

PR #578, opened 2016-02-21
fix arguments priority parsing when thy have no priority

agallou:
When an extension registered a new command line option,
and when the config file the loads this extension is loaded
via the command line, we could have issues depending on the order
of the arguments.

For example, this were not working:

```
 ./bin/atoum -c config/.atoum.php --autoloop
```

(we get this error `Error: Argument '--autoloop' is unknown, did you mean '--help'?`).

But this were working:

```
 ./bin/atoum --autoloop  -c config/.atoum.php
```

That because the --autoloop argument is not is the priorities array
beacause it's still not loaded.

So, in order to fix that, we always put the arguments with no priority
add then end.

Like that, other arguments, like adding a config file, that have the PHP_INT_MAX
value (the maximum priority), will be loaded before the arguments with no priority.
Then the extension will be loaded the the --autoloop argument will exist, and we will
not get an error.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,10 @@
 * [#569](https://github.com/atoum/atoum/pull/569) Use in-memory cache for resolved asserters ([@jubianchi])
 * [#567](https://github.com/atoum/atoum/pull/567) Extract loop logic from runner and add a looper interface ([@jubianchi], [@agallou])
 
+## Bugfix
+
+* [#578](https://github.com/atoum/atoum/pull/578) Fix arguments priority parsing when they have no priority ([@agallou])
+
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
 # `dev-master`
 
+* [#578](https://github.com/atoum/atoum/pull/578) Fix arguments priority parsing when they have no priority ([@agallou])
```
jubianchi: Please, can you add this to a Bugfix section (example here: https://github.com/atoum/atoum/blob/master/CHANGELOG.md#bugfix)

agallou: @jubianchi done

PR #585, opened 2016-03-02
Memory usage is based on the peak & real allocation

Hywan:
The memory usage was previously based on the system memory. By system, PHP means the VM. So first, we should use the real allocated memory. Second, we use `memory_get_peak_usage` instead of `memory_get_usage` because the GC (Garbage Collector) can run and free the memory. Thus, a test case can report 0 byte while it has allocated 1Mb for instance. The peak is what is interesting for us, and the allocated peak.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 # `dev-master`
 
+* [#585](https://github.com/atoum/atoum/pull/585) Memory usage is based on the peak & real allocations ([@hywan])
 * [#740](https://github.com/atoum/atoum/pull/740) String asserter now has `notMatches` assertion ([@fvilpoix])
 
 # 3.2.0 - 2017-09-07
--- a/classes/test.php
+++ b/classes/test.php
@@ -1264,7 +1264,7 @@ abstract class test implements observable, \countable
 
                     $assertionNumber = $this->score->getAssertionNumber();
                     $time = microtime(true);
-                    $memory = memory_get_usage(true);
+                    $memory = memory_get_peak_usage();
 
                     if (isset($this->dataProviders[$testMethod]) === false) {
                         $this->{$testMethod}();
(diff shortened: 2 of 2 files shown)


mikaelrandy:
:+1:

jubianchi:
@Hywan please can you rebase this branch and add a changelog entry ? I would like to get this merged before the next release.

Thanks.

mikaelrandy:
@Hywan i will release 2.7.0 tomorrow.
It seems this PR is ready to merge as soon as you will rebase it.

jubianchi:
@Hywan I checked out your branch and implemented a dedicated field for peak memory usage. If I pass `true` as the `$real_usage` of `memory_get_peak_usage` I get the exact same result as `memory_get_usage(true)`.

If i leave the default value (i.e `memory_get_peak_usage(false)`) I get a lower result, for example:

```
=> Memory usage: 9.75 Mb.
=> Peak memory usage: 8.32 Mb.
```

What do you expect from this patch ?

jubianchi:
@Hywan ping ?

Hywan:
@jubianchi If you get the same results from `memory_get_usage` and `memory_get_peak_usage`, it means the GC did not run yet. It can happen on small test case. If the “real usage” parameter is set to its default value (aka `false`), then only the used memory is reported, which excludes all the unused pages. This is kind of the difference between “reserved/allocated” (real usage = `true`) and “used” (real usage = `false`). We do not really care about the amount of allocated here, we are more interested by the actual memory consumption. If the allocator reserved more memory than necessary, I am not sure we are concerned about it. However, we are concerned by the amount of memory that is actually used (not empty). Thoughts?

Grummfy:
perhaps for v3?

Hywan:
If we do not intent to release any versions until the 3.0, then yes.

Grummfy:
what about this one?

jubianchi:
good for me

Grummfy:
youhou

Issue #586, opened 2016-03-03
PHAR link in README is broken


personman:
The link points to http://downloads.atoum.org/nightly/atoum.phar but that's 404'd.

Maybe this is the replacement?

http://downloads.atoum.org/nightly/mageekguy.atoum.phar

CedCannes:
+1 PHAR link is broken 

the link http://downloads.atoum.org/nightly/mageekguy.atoum.phar points a very old version of the PHAR.

@Grummfy gave me the link to the latest version : 
https://github.com/atoum/atoum/releases/tag/2.6.0
https://github.com/atoum/atoum/releases/download/2.6.0/atoum.phar

Following the README.md and the official documentation I have not managed to find the latest version.

PR #588, opened 2016-03-16
make PHAR link point to a recent version

CedCannes:
I followed the old link pointing to download.atoum.org that contains version : nightly - 2416-201402121146 of the PHAR and I couldn't use atoum correctly with the new features listed in her pretty documentation. 
I think it is detrimental for those who want to discover atoum
as mentioned in #586

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@ Finally, even though it is developed mainly on UNIX, it can also work on Windows
 
 ## Why atoum?
 
-* *atoum* is really [easy to install](http://docs.atoum.org/en/latest/getting_started.html#installation): clone it [from github](https://github.com/atoum/atoum), download [its PHAR](http://downloads.atoum.org/) or simply [use composer](https://packagist.org/packages/atoum/atoum),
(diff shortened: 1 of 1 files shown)


Hywan:
Why can't we use `latest` instead of `2.6.0` in the URL?

jubianchi:
@Hywan download URLs are not allowed to use the `latest` word :/

Grummfy:
This is why I have made a script a long time ago to get the last version.

For me this issue is the reflect of what I insist since more than a year to fix this issue. A simple redirection will solve the problem...

mageekguy:
Hi folks,

I think that the problem is… me ;)
It seems to me that in galaxy far far far away, someone send to me the URL which will be used to redirect to the right version of the PHAR, but i have forget to do the redirection…
So, if someone send me (again) the URL… maybe i will not forget to configure the legacy web server accordingly!

Hywan:
cc @Grummfy

Grummfy:
I have no link ;) I have create a file to redirect people and say that we probably need a cache or something like that....

https://gist.github.com/Grummfy/2f040c293dce5109810f

but that's all. So if you want I can create a better script with some cache and give it to mageeguy to do the job, if t's ok for you.

PR #596, opened 2016-05-06
Test methods' tags are inherited from test classes

### LINKED ISSUE #573: `@tag`s are not inherited from class to method definition
`@tag`s are not inherited from class definition to method definition. So:
```php
/* @tag foo */
class C {
    /* @tag bar */
    public function f() { … }
}
```
`f` has the tag `bar`, not `foo` **and** `bar`.

jubianchi:
Closes #573


Grummfy:
seems ok, no?

PR #600, opened 2016-05-19
Reports override correctly when using -ulr/-utr

jubianchi:
Before this patch, if we had a configuration file like this one:

```php
<?php

$script->addDefaultReport();

$report = new some\report();
//...
$runner->addReport($telemetry);
```

And used the `-utr` or `-ulr` options, atoum would have dropped the additional report we added through the configuration file and only used the light CLI report or TAP report.

Now, when we do that, atoum will only remove reports which are conflicting with light or TAP report and keep the others.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ## Bugfix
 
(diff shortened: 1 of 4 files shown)


mikaelrandy:
I will release 2.7.0 tomorrow.
I think this bugfix should be included in.
Are you ready to merge ?

jubianchi:
let's go :)

Issue #552, opened 2016-01-08
add stubs in the require section of composer.json instead of suggest


agallou:
One of the main complaints we heard about atoum was the lack of auto-completion.

Requering the atoum/stubs package fixes this problem.

But it's not a great user experience. You have to read the documentation or the composer output do that : some people still does'nt now about this package (If we look on packagist, on the 30 last days atoum/atoum has been installed 6 287 times. And the stubs package 1 025 times)

How about putting the stubs package as a requirement for atoum ? (for now it's on the suggestion section of the composer.json).

With this : 
* the user will have out of the box completion/documentation. he won't have to do anything.
* the phpdoc will still be outside atoum's repository

What do you think about this ?

jubianchi:
atoum will stay standalone. There should be nothing in the require section. 

We'll keep stubs in the suggests. Sorry.

PR #604, opened 2016-06-05
Add a way to define configuration in atoum/*-edition packages

agallou:
This will allow us to create "editions" of atoum, will extensions preloaded and some stuff configured.

For example : https://github.com/agallou/atoum-standard-edition

By requirering only atoum/standard-edition this will : 
* install atoum/atoum
* install atoum/stubs
* install atoum/autoloop-extension
* install atoum/visibility-extension
* install atoum/config-extension
* install atoum/reports-extension
* load autoloop-extension
* load visibility-extension
* load config-extension
* load reports-extension
* if the TELEMETRY_ENABLED environement variable is defined, the telemetry report will be sent (with different anonymization settings depending of it's value (1, 2 ou 3). 

we could also imagine
*  preconfiguring the coveralls report repending on an COVERALLS_API_KEY environnement variable
* configure the autoloop using the files defined in the autoload from the project's composer.json
* add a XUNIT_REPORT_PATH : that will generate the xunit report in the path specified in the environnement variable : that will also help the integration in CI environments.

you can see an example of usage here : https://github.com/agallou/atoum-std-edition-example (the repositories and atoum/atoum in the require section should not be here. They won't be if diffrent PR are merged).

This standard edition will help users set-up atoum.
This will also : 
* have stubs installed with just one require (we can see the advantages here https://github.com/atoum/atoum/issues/552 : out of the box autocompletion and no phpdoc in the atoum core)
* have the configuration extension preinstalled : users that don't like the configuration in PHP wil be able to set some configuration with a YAML file (for example https://github.com/agallou/atoum-std-edition-example/blob/master/.atoum.yml ). Users that prefers the .atoum.php file will still continue to be able to use it as usual (with all i'ts benefits (configuration in php, we can do everything we want)).

For now we still need to define manually a .bootstrap.atoum.php file, but if the PR https://github.com/atoum/atoum/pull/523 is also merged, the installation/configuration will even be easier.

### DIFF
--- a/classes/runner.php
+++ b/classes/runner.php
@@ -835,14 +835,27 @@ class runner implements observable
 
 	public function removeExtension(extension $extension)
 	{
-		$this->extensions->detach($extension);
+		try {
+			$extension = $this->getExtension(get_class($extension));
 
-		return $this->removeObserver($extension);
+			$this->extensions->detach($extension);
+
+			$this->removeObserver($extension);
+		} catch (exceptions\logic\invalidArgument $e) {
+		}
+
+		return $this;
 	}
 
 	public function removeExtensionByClassName($classname)
 	{
-		return $this->removeExtension($this->getExtension($classname));
+		try {
+			$this->removeExtension($this->getExtension($classname));
+		} catch (exceptions\logic\invalidArgument $e) {
+		}
+
+		return $this;
+
 	}
 
 	public function removeExtensions()
--- a/classes/runner.php
+++ b/classes/runner.php
@@ -867,6 +867,27 @@ class runner implements observable
 		return $this;
 	}
 
+	public function getExtension($classname)
+	{
+		$extensions = array();
+		foreach ($this->getExtensions() as $extension)
+		{
+			if (get_class($extension) != $classname)
+			{
+				continue;
+			}
+
+			$extensions[] = $extension;
+		}
+
+		if (1 != count($extensions))
+		{
+			throw new exceptions\logic\invalidArgument(sprintf("Extension '%s' not found", $classname));
+		}
+
+		return array_shift($extensions);
+	}
+
 	public static function isIgnored(test $test, array $namespaces, array $tags)
 	{
 		$isIgnored = $test->isIgnored();
--- a/classes/runner.php
+++ b/classes/runner.php
@@ -59,7 +59,7 @@ class runner implements observable
 
 		$this->observers = new \splObjectStorage();
 		$this->reports = new \splObjectStorage();
-		$this->extensions = new \splObjectStorage();
+		$this->extensions = new iterators\classStorage();
 	}
 
 	public function setAdapter(adapter $adapter = null)
@@ -869,7 +869,6 @@ class runner implements observable
 
 	public function getExtension($classname)
 	{
-		$extensions = array();
 		foreach ($this->getExtensions() as $extension)
 		{
 			if (get_class($extension) != $classname)
@@ -877,15 +876,10 @@ class runner implements observable
 				continue;
 			}
 
-			$extensions[] = $extension;
+			return $extension;
 		}
 
-		if (1 != count($extensions))
-		{
-			throw new exceptions\logic\invalidArgument(sprintf("Extension '%s' not found", $classname));
-		}
-
-		return array_shift($extensions);
+		throw new exceptions\logic\invalidArgument(sprintf("Extension '%s' not found", $classname));
 	}
 
 	public static function isIgnored(test $test, array $namespaces, array $tags)
--- a/classes/runner.php
+++ b/classes/runner.php
@@ -840,6 +840,11 @@ class runner implements observable
 		return $this->removeObserver($extension);
 	}
 
+	public function removeExtensionByClassName($classname)
+	{
+		return $this->removeExtension($this->getExtension($classname));
+	}
+
 	public function removeExtensions()
 	{
 		foreach ($this->extensions as $extension)
--- a/classes/scripts/runner.php
+++ b/classes/scripts/runner.php
@@ -33,6 +33,7 @@ class runner extends atoum\script\configurable
 
 	protected static $autorunner = true;
 	protected static $runnerFile = null;
+	protected static $configurationCallables = array();
 
 	public function __construct($name, atoum\adapter $adapter = null, atoum\scripts\runner\looper $looper = null)
 	{
@@ -245,6 +246,20 @@ class runner extends atoum\script\configurable
 		$runner = $this->runner;
 
 		return $this->includeConfigFile($path, function($path) use ($script, $runner) { include_once($path); });
+	}
+
+	public function useConfigCallable($callback)
+	{
+		if (!(is_callable($callback) && $callback instanceof \Closure) ){
+			throw new exceptions\logic\invalidArgument('The argument must be a closure');
+		}
+
+		$script = call_user_func($this->configuratorFactory, $this);
+		$runner = $this->runner;
+
+		$callback($script, $runner);
+
+		return $this;
 	}
 
 	public function testIt()
@@ -683,12 +698,21 @@ class runner extends atoum\script\configurable
 
 		static::$autorunner = new static($name);
 
+		foreach (static::$configurationCallables as $callable) {
+			static::$autorunner->useConfigCallable($callable);
+		}
+
 		return static::$autorunner;
 	}
 
 	public static function disableAutorun()
 	{
 		static::$autorunner = false;
+	}
+
+	public static function addConfigurationCallable($callable)
+	{
+		static::$configurationCallables[] = $callable;
 	}
 
 	protected function setArgumentHandlers()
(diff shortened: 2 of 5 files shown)


Grummfy:
thanks

agallou:
https://github.com/agallou/atoum-standard-edition has been updated to ease the configuration of the xUnit report. 

With this, for example, the configuration in Jenkins will be a lot easier. No config file will be needed.

only an environnement variable will be needed : 
```
export XUNIT_REPORT_PATH=app/logs/xunit.xml
./vendor/bin/atoum                         
```

and to xml file will be created (so as the directory if it does not) : 
```
tree app 
app
└── logs
    └── xunit.xml

1 directory, 1 file
```

jubianchi:
This solution makes atoum coupled to eventual editions: I'm not really comfortable with that.

Here what I would like to propose as an alternative: edition would push their configuration to atoum using autoloader. Here is how.

The extension will have to declare an autoloader file:

```json
{
    ...
    "autoloaders: {
        "files": [
            "configuration.php" 
        ]
    }
}
```

Inside this `configuration.php` file there will be the code used to pre-configure atoum:

```php
<?php

use mageekguy\atoum;
use mageekguy\atoum\scripts;

if (defined('mageekguy\atoum\scripts\runner') === true) {
    scripts\runner::configureRunner(function(atoum\runner $runner) {
        $runner->addExtension(new atoum\blackfire\extension());
        //...
    });
}
```

Then in the project's `.atoum.php` file we can imagine something like:

```php
<?php

$runner->getExtension(atoum\blackfire\extension::class)
    -> setClientConfiguration(/*...*/)
;
```

This would require less coupling but some work has to be done in atoum:

* [ ] Add the `mageekguy\atoum\scripts\runner::configureRunner` method
* [ ] Add the `mageekguy\atoum\runner::getExtension` method

What do you think ?

agallou:
The getExtension could be very usefull. I will add this.

But using composer's autoloader will be problematic (i've also tried it when working on this PR).

The issue there is that atoum does not uses (yet?) composer's autoloader.

Samll change on your proposal, instead of passing a callback, i've tested by adding an array of files that will by loaded by the enableAutorun method : 

Like this :
```
+++ b/classes/scripts/runner.php
@@ -29,8 +29,10 @@ class runner extends atoum\script\configurable
        protected $loop = false;
        protected $looper;
 
+
        protected static $autorunner = true;
        protected static $runnerFile = null;
+       protected static $autorunnerConfigurationFiles = array();
 
        public function __construct($name, atoum\adapter $adapter = null, atoum\scripts\runner\looper $looper = null)
        {
@@ -644,6 +646,10 @@ class runner extends atoum\script\configurable
 
                static::$autorunner = new static($name);
 
+               foreach (static::$autorunnerConfigurationFiles as $file) {
+                       static::$autorunner->useConfigFile($file);
+               }
+
                return static::$autorunner;
        }
 
@@ -652,6 +658,11 @@ class runner extends atoum\script\configurable
                static::$autorunner = false;
        }
 
+       public static function addAutorunnerConfigurationFile($file)
+       {
+               static::$autorunnerConfigurationFiles[] = $file;
+       }
```

the standard edition file has be changed to add an autoloader containing this : 
```
<?php

use mageekguy\atoum;
use mageekguy\atoum\scripts;

if (defined('mageekguy\atoum\scripts\runner') === true) {
    scripts\runner::addAutorunnerConfigurationFile(__DIR__ . DIRECTORY_SEPARATOR . 'configuration.php');
}
```

This allows us to have configuration files similar to the .atoum.php ones.

By it doesn't work as is.

In order to make this work we need to add 
```
require_once __DIR__ . '/../../../autoload.php';
```
in the `vendor/atoum/atoum/scripts/runner.php` file.

(or apply the patch of the #523 PR)

Without that, atoum will never load composer's autoloader, and the configuration will not be loaded.

So, about the autoloader : 
* if we require composer's autoloader inside atoum, we will be able to use an autoload file to configure the autorunner
* if we don't require composer's autoloader, we will have to find another way to automatically load the configuration (like the way it's done for the moment in the PR). But like you said, this ties atoum with the editions.

What's your thought's about that ?
(and if we go for requiring composer's autoloader, what can we do to merge the PR #523  ?)

classes/scripts/runner.php
```diff
@@ -652,6 +657,11 @@ public static function disableAutorun()
 		static::$autorunner = false;
 	}
 
+	public static function addAutorunnerConfigurationFile($file)
```
jubianchi: `addConfigurationFile`

classes/scripts/runner.php
```diff
@@ -31,6 +31,7 @@ class runner extends atoum\script\configurable
 
 	protected static $autorunner = true;
 	protected static $runnerFile = null;
+	protected static $autorunnerConfigurationFiles = array();
```
jubianchi: `$configurationFiles`

classes/runner.php
```diff
@@ -844,6 +844,23 @@ public function addExtension(extension $extension, extension\configuration $conf
...
+		$extensions = array();
+		foreach ($this->getExtensions() as $extension) {
+			if (get_class($extension) != $classname) {
+				continue;
+			}
+			$extensions[] = $extension;
+		}
+
+		if (1 != count($extensions)) {
+			throw new \Exception(sprintf("Extension '%s' Not found", $classname));
```
jubianchi: `s/Not/not/`

jubianchi: @agallou use a more specific exception, [https://github.com/atoum/atoum/blob/master/classes/exceptions/logic/invalidArgument.php] for example ;)

classes/runner.php
```diff
@@ -844,6 +844,23 @@ public function addExtension(extension $extension, extension\configuration $conf
 		return $this;
 	}
 
+	public function getExtension($classname)
+	{
+		$extensions = array();
+		foreach ($this->getExtensions() as $extension) {
```
jubianchi: `{` should be on its on line ;)

classes/runner.php
```diff
@@ -844,6 +844,23 @@ public function addExtension(extension $extension, extension\configuration $conf
...
+			if (get_class($extension) != $classname) {
```
jubianchi: `{` should be on its on line ;)

classes/runner.php
```diff
@@ -844,6 +844,23 @@ public function addExtension(extension $extension, extension\configuration $conf
...
+				continue;
+			}
+			$extensions[] = $extension;
```
jubianchi: could you add a blank line before this line

classes/scripts/runner.php
```diff
@@ -244,6 +245,16 @@ public function useConfigFile($path)
 		return $this->includeConfigFile($path, function($path) use ($script, $runner) { include_once($path); });
 	}
 
+	public function useConfigCallable(callable $callback)
```
jubianchi: `useConfigurationCallable`

classes/scripts/runner.php
```diff
@@ -244,6 +245,16 @@ public function useConfigFile($path)
...
+	{
+		$script = call_user_func($this->configuratorFactory, $this);
```
jubianchi: `$script` should be `$this` ?

agallou: @jubianchi i've get it the same way as in useConfigFile : 
```
public function useConfigFile($path)
 	{
 		$script = call_user_func($this->configuratorFactory, $this);
 		$runner = $this->runner;
 
 		return $this->includeConfigFile($path, function($path) use ($script, $runner) { include_once($path); });
 	}
```

classes/scripts/runner.php
agallou: but, there a method called useConfigFile. would'nt it be consistent to name it the same way ?

classes/scripts/runner.php
jubianchi: ok fine ;)

classes/scripts/runner.php
jubianchi: fine ;)

classes/runner.php
```diff
@@ -844,6 +844,27 @@ public function addExtension(extension $extension, extension\configuration $conf
...
+			{
+				continue;
+			}
+
+			$extensions[] = $extension;
+		}
+
+		if (1 != count($extensions))
+		{
+			throw new \InvalidArgumentException(sprintf("Extension '%s' not found", $classname));
```
jubianchi: please, can you throw the `mageekguy\atoum\exceptions\logic\invalidArgument`

classes/runner.php
```diff
@@ -844,6 +844,27 @@ public function addExtension(extension $extension, extension\configuration $conf
...
+		{
+			if (get_class($extension) != $classname)
+			{
+				continue;
+			}
+
+			$extensions[] = $extension;
+		}
+
+		if (1 != count($extensions))
```
jubianchi: I think we should improve the `addExtension` method and move this check there: atoum should refuse to load an extension twice.

The `addExtension` already checks for an already existing extension instance but it should also check if an extension of the same class is already loaded.

Doing this, we could make the code in `getExtension` simpler: return early in the foreach loop and throw if we exit the loop without any matching extension.

jubianchi:
One last comment in the code.

Would you mind adding the `removeExtension` method in this PR ?

agallou:
@jubianchi PR updated : 
* the extension storage now only supports one extension of the same class (if another extension with the same class is passed, it will be ignore just like if we put the same object right now)
* the getExtension is indeed now simpler
* the removeExtension already exists  a method to remove it from it's class name has been added (but wil the same limitations has the existing method : the asserters will be removed but not the arguments)

jubianchi:
I'm closing this one in favor of #634 and #635.

Thanks for your work @agallou!

PR #605, opened 2016-06-19
Automatically include Composer's autoloader if it exists

jubianchi:
Handle `.autoloader.atoum.php` files to define tests autoloader: if present this file will be automatically loaded by atoum. It can be useful if the user has to include several autoloaders or do not use composer.

If `.autoloader.atoum.php` does not exist, atoum will try to find the Composer autoloader searching from the current working directory up to the FS root for `vendor/autoload.php`.

Add the `--autoloader-file`/`-af` CLI argument to define which autoloader file to user.

Closes #523

### DIFF
--- a/classes/scripts/runner.php
+++ b/classes/scripts/runner.php
@@ -16,6 +16,8 @@ class runner extends atoum\script\configurable
 {
 	const defaultConfigFile = '.atoum.php';
 	const defaultBootstrapFile = '.bootstrap.atoum.php';
+	const defaultAutoloaderFile = '.autoloader.atoum.php';
+	const defaultComposerAutoloaderFile = 'vendor/autoload.php';
 
 	protected $runner = null;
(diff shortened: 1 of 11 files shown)


classes/scripts/runner.php
```diff
@@ -16,6 +16,8 @@ class runner extends atoum\script\configurable
 {
 	const defaultConfigFile = '.atoum.php';
 	const defaultBootstrapFile = '.bootstrap.atoum.php';
+	const defaultAutoloaderFile = '.autoloader.atoum.php';
+	const defaultComposerAutoloaderFile = 'vendor/autoload.php';
```
agallou: what happens if we run atoum from a directory that does not contains vendor ? (for example, il we launch if from the tests dir with `../vendor/bin/atoum` ? does the includer will find the autoload ?

Hywan:
:+1: 🎉

PR #612, opened 2016-07-07
Add telemetry support to atoumtask

oxman:
Add telemetry support to AtoumTask for Phing

### DIFF
--- a/resources/phing/AtoumTask.php
+++ b/resources/phing/AtoumTask.php
@@ -103,10 +103,6 @@ class atoumTask extends task
 			throw new exception('AtoumTask depends on Xdebug being installed to gather code coverage information');
 		}
 
-		if ($this->telemetry && class_exists('\mageekguy\atoum\reports\telemetry') === false) {
-			throw new exception('AtoumTask depends on atoum/reports-extension being installed to enable telemetry report');
-		}
-
 		if ($this->atoumPharPath !== null)
 		{
 			require_once $this->atoumPharPath;
@@ -184,10 +180,14 @@ class atoumTask extends task
 
 		if ($this->telemetryEnabled())
 		{
+			if (class_exists('\mageekguy\atoum\reports\telemetry') === false) {
+				throw new exception('AtoumTask depends on atoum/reports-extension being installed to enable telemetry report');
+			}
+
 			$telemetry = new \mageekguy\atoum\reports\telemetry();
 			$telemetry->addWriter(new \mageekguy\atoum\writers\std\out());
 
-			if ($this->getTelemetryProjectName() !== '') {
+			if ($this->getTelemetryProjectName() !== null) {
 				$telemetry->setProjectName($this->getTelemetryProjectName());
 			}
 			$runner->addReport($telemetry);
--- a/resources/phing/AtoumTask.php
+++ b/resources/phing/AtoumTask.php
@@ -180,12 +180,12 @@ class atoumTask extends task
 
 		if ($this->telemetryEnabled())
 		{
-			if (class_exists('\mageekguy\atoum\reports\telemetry') === false) {
+			if (class_exists('reports\telemetry') === false) {
 				throw new exception('AtoumTask depends on atoum/reports-extension being installed to enable telemetry report');
 			}
 
-			$telemetry = new \mageekguy\atoum\reports\telemetry();
-			$telemetry->addWriter(new \mageekguy\atoum\writers\std\out());
+			$telemetry = new reports\telemetry();
+			$telemetry->addWriter(new atoum\writers\std\out());
 
 			if ($this->getTelemetryProjectName() !== null) {
 				$telemetry->setProjectName($this->getTelemetryProjectName());
--- a/resources/phing/AtoumTask.php
+++ b/resources/phing/AtoumTask.php
@@ -26,6 +26,8 @@ class atoumTask extends task
 	private $codeCoverageTreemapUrl = null;
 	private $codeCoverageXunitPath = null;
 	private $codeCoverageCloverPath = null;
+	private $telemetry = false;
+	private $telemetryProjectName = null;
 	private $atoumPharPath = null;
 	private $atoumAutoloaderPath = null;
 	private $phpPath = null;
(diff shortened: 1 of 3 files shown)


Grummfy:
Thanks for the PR.

We should probably add a check to be sure that atoum report extension is available before ?

Otherwise seems good.

oxman:
Already done : https://github.com/atoum/atoum/pull/612/files#diff-ab8a2832f2790fe82b1aca7e5afb66f7R183

resources/phing/AtoumTask.php
```diff
@@ -171,6 +178,21 @@ public function execute()
 			}
 		}
 
+		if ($this->telemetryEnabled())
+		{
+			if (class_exists('\mageekguy\atoum\reports\telemetry') === false) {
+				throw new exception('AtoumTask depends on atoum/reports-extension being installed to enable telemetry report');
```
jubianchi: I would not thrown an exception here to make the dependency optional.

resources/phing/AtoumTask.php
```diff
@@ -171,6 +178,21 @@ public function execute()
...
+			}
+
+			$telemetry = new \mageekguy\atoum\reports\telemetry();
```
jubianchi: this whole block should be in the `if`

resources/phing/AtoumTask.php
oxman: The dependency is not optional if you enabled telemetry, its a mandatory.  
But if you don't enable telemetry the dependency is optional.

resources/phing/AtoumTask.php
oxman: Why ?

resources/phing/AtoumTask.php
jubianchi: you are totally right I misread your code

resources/phing/AtoumTask.php
jubianchi: nevermind

Grummfy:
@jubianchi now it's seems ok, so we can merge?

resources/phing/AtoumTask.php
jubianchi:
```diff
- AtoumTask depends on atoum/reports-extension being installed to enable telemetry report
+ AtoumTask requires atoum/reports-extension to enable telemetry report
```

resources/phing/AtoumTask.php
jubianchi: @oxman could you use the short name for this class: `reports\telemetry` ? We already have the required `use` statement.

resources/phing/AtoumTask.php
```diff
@@ -171,6 +178,21 @@ public function execute()
...
+			$telemetry->addWriter(new \mageekguy\atoum\writers\std\out());
```
jubianchi: same thing here: `atoum\writers\std\out`

resources/phing/AtoumTask.php
```diff
@@ -287,13 +309,37 @@ public function getCodeCoverage()
 		return $this->codeCoverage;
 	}
 
+	public function setTelemetry($telemetry)
```
jubianchi: I would have split this into two method: 
* `enableTelemetry()`
* `disableTelemetry()`

resources/phing/AtoumTask.php
```diff
@@ -287,13 +309,37 @@ public function getCodeCoverage()
...
+	{
+		$this->telemetry = (boolean) $telemetry;
+
+		return $this;
+	}
+
+	public function getTelemetry()
```
jubianchi: `TelemetryIsEnabled()` to be consistent with what we do in other class (e.g https://github.com/atoum/atoum/blob/master/classes/test.php#L719)

tests/units/resources/phing/AtoumTask.php
```diff
@@ -543,5 +547,25 @@ public function testSetCodeCoverageCloverPath()
...
 			;
 		}
+
+		public function testGetSetTelemetry()
+		{
+			$this
+				->given($task = new testedClass())
+				->then
+					->object($task->setTelemetry(true))->isIdenticalTo($task)
+					->boolean($task->getTelemetry())->isEqualTo(true)
```
jubianchi:
```diff
- ->isEqualTo(true)
- ->isTrue
```

tests/units/resources/phing/AtoumTask.php
```diff
@@ -543,5 +547,25 @@ public function testSetCodeCoverageCloverPath()
...
+			;
+		}
+
+		public function testGetSetTelemetryProjectName()
+		{
+			$this
+				->given($task = new testedClass())
+				->then
+				->object($task->setTelemetryProjectName($projectName = uniqid() . '/' . uniqid()))->isIdenticalTo($task)
```
jubianchi: this line should be indented

tests/units/resources/phing/AtoumTask.php
```diff
@@ -543,5 +547,25 @@ public function testSetCodeCoverageCloverPath()
...
+				->string($task->getTelemetryProjectName())->isEqualTo($projectName)
```
jubianchi: this line should be indented

tests/units/resources/phing/AtoumTask.php
```diff
@@ -543,5 +547,25 @@ public function testSetCodeCoverageCloverPath()
...
 		}
+
+		public function testGetSetTelemetry()
+		{
+			$this
+				->given($task = new testedClass())
+				->then
+					->object($task->setTelemetry(true))->isIdenticalTo($task)
+					->boolean($task->getTelemetry())->isEqualTo(true)
+			;
```
jubianchi: you are not testing the case where we disable the telemetry

jubianchi:
@oxman some things left to do:

* I reviewed some bits of the code, could you fix it please?
* Could you also add an entry to the changelog pointing to your PR with credits to you
* Could you please sign your commit (optional but would be appreciated)

resources/phing/AtoumTask.php
oxman: Not conform to other test

resources/phing/AtoumTask.php
oxman: Not conform to other test in this file

PR #614, opened 2016-07-08
Ability to register additional error handlers for tests

lolautruche:
This patch makes it possible to add one or several error handlers for a test.
This is particularly useful when one wants to deal with silenced errors.

One use case is to deal with Symfony `E_USER_DEPRECATED`, which are silenced on purpose.
Those errors are then caught by the debug bar.

With this patch, one will be able to implement an extension and add one or several error handlers.
My idea is to implement the equivalent of Symfony PHPUnitBridge which would catch deprecation errors.

### DIFF
--- a/classes/test.php
+++ b/classes/test.php
@@ -78,6 +78,7 @@ abstract class test implements observable, \countable
 	private $classHasNotVoidMethods = false;
 	private $extensions = null;
 	private $analyzer;
+	private $errorHandlers = array();
 
 	private static $namespace = null;
 	private static $methodPrefix = null;
@@ -1517,8 +1518,23 @@ abstract class test implements observable, \countable
 
 			$doNotCallDefaultErrorHandler = !($errno & E_RECOVERABLE_ERROR);
 		}
+		
+		foreach ($this->errorHandlers as $handler)
+		{
+			call_user_func($handler, $errno, $errstr, $errfile, $errline);
+		}
 
 		return $doNotCallDefaultErrorHandler;
+	}
+	
+	public function registerErrorHandler($errorHandler)
+	{
+		if (!is_callable($errorHandler))
+		{
+			throw new exceptions\logic\invalidArgument('Error handler is not callable');
+		}
+
+		$this->errorHandlers[] = $errorHandler;
 	}
 
 	public function setUp() {}


jubianchi:
@lolautruche did you saw [this extension](https://github.com/atoum/deprecated-extension) ?

atoum also provides a dedicated asserter for errors: http://docs.atoum.org/en/latest/asserters.html#error

lolautruche:
Thanks @jubianchi for linking me this extension. This is not exactly what I want to do, but I think I can contribute something with this patch in. Basically we could catch silenced deprecated errors directly, without having to traverse the entire score... \o/

lolautruche:
All green now \o/

mikaelrandy:
👍

jubianchi:
@lolautruche I took some time to review this PR and there is one issue with it. Given the following class and associated test:

```php
<?php

namespace
{
    class foo
    {
        public function bar()
        {
            trigger_error(__METHOD__, E_USER_DEPRECATED);
        }
    }
}

namespace tests\units
{
    use atoum;

    /**
     * @engine inline
     */
    class foo extends atoum
    {
        public function testFoo()
        {
            $this->registerErrorHandler(function($errno) {
                var_dump($errno);
            });

            $this->newTestedInstance;
            $this->testedInstance->bar();

            $this->boolean(true);
        }

        public function testBar()
        {
            $this->newTestedInstance;
            $this->testedInstance->bar();

            $this->boolean(true);
        }
    }
}

```

When running this test, we expect to only have an error handler for the first method (`testFoo`) but it applies to both method. Why? Because when we use the `inline` engine, we do not have any isolation thus the test class is "recycled" for each test method. To fix this issue, we have to reset error handlers for each test method [here](https://github.com/atoum/atoum/blob/master/classes/test.php#L1194).

Once this is done, we can consider this PR is done. I'm looking at how we can use this new feature in extension (for example, in the `deprectaed-extension`) but is harder than it seems, especially when we have isolation.

lolautruche:
Hi. I get it and will update the PR accordingly, thanks. 
As for the extension, I already made a few tests and it's indeed tricky to keep trace of depreciation, be cause of sub-processes. However, the problem is the same with phpunit and seems to be solved by using static methods and static variables. 
I'm pretty sure we can solve this :wink:

jubianchi:
@lolautruche I have it solved on my local repository ;) The code is a bit dirty but I can try to make it clean and push it for test asap if you wish ;)

jubianchi:
ping @lolautruche, some questions:

* do you still need this feature? AFAIR, perhaps you only need to use the [atoum/deprecated-extension](https://github.com/atoum/deprecated-extension).
* did you manage to fix the problem I mentionned [here](https://github.com/atoum/atoum/pull/614#issuecomment-231864045)

lolautruche:
@jubianchi 

> do you still need this feature? AFAIR, perhaps you only need to use the atoum/deprecated-extension.

Yes I do. As I mentioned at the beginning:

> Thanks @jubianchi for linking me this extension. This is not exactly what I want to do, but I think I can contribute something with this patch in.

> did you manage to fix the problem I mentionned here

Actually I dropped everything during my summer vacation, so not yet :-). But I know I'll need this soon.

I'll dive into it ASAP.

jubianchi:
@lolautruche perfect! I'll remove this from the 2.9.0 milestone for now. If you update the PR in time for the release I'll include it. But for now, I'll keep it for a future release.

Thanks!

Grummfy:
hello I pop a bit this issue to see if: it's still required and if we can help to finalize it. As I understand it's just a small fix to make it ready.

Issue #598, opened 2016-05-10
Remove reserved keyword for PHP-7.1.0


vonglasow:
Apparently `void` should be a reserved keyword in PHP-7.1.0

example: https://continuousphp.com/git-hub/vonglasow/praspel/build/c6c9e55e-3609-4977-bfd4-c78343451277/output

=> see output for the version PHP-7.1.0
To check with the branch master of PHP. (No tag found)

remicollet:
Indeed:

```
+ php71 runner.php --directories .
Fatal error: Cannot use 'void' as class name as it is reserved in /dev/shm/BUILD/atoum-a3bceadf39cccd4cc321407eeb61187bdee438cd/classes/report/fields/runner/tests/void.php on line 11

```

Hywan:
Official source: https://github.com/php/php-src/blob/php-7.1.0alpha2/UPGRADING.

vonglasow:
To complete this issue:
- [ ] replace void by blank
- [ ] complete changelog
- [ ] add alias in autoloading

Hywan:
Please, update the `CHANGELOG` and we are good :-).

Hywan:
It's like we are good now. Please @vonglasow, can you write the `CHANGELOG` entry?

vonglasow:
CHANGELOG entry added!

mvrhov:
When will this get merged? I have an php 7.1 only application and I'm unable to run tests.

jubianchi:
#615 was merge. 3.0.0 release will be coming soon!

PR #615, opened 2016-07-10
Replace void by blank

vonglasow:
Void is now a reserved keyword in PHP-7.1. This patch replace all
required call from `void` to `blank`

This PR fix https://github.com/atoum/atoum/issues/598

### DIFF
--- a/classes/autoloader.php
+++ b/classes/autoloader.php
@@ -41,7 +41,12 @@ class autoloader
 			$this->addNamespaceAlias($alias, $target);
 		}
 
-		foreach ($classAliases ?: array('atoum' => __NAMESPACE__ . '\test', __NAMESPACE__ => __NAMESPACE__ . '\test') as $alias => $target)
+		$defaultAliases = array(
+			'atoum' => __NAMESPACE__ . '\test',
+			__NAMESPACE__ => __NAMESPACE__ . '\test'
+		);
+
+		foreach ($classAliases ?: $defaultAliases as $alias => $target)
 		{
 			$this->addClassAlias($alias, $target);
 		}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 # `dev-master`
 
+* [#615](https://github.com/atoum/atoum/pull/615) Remove reserved keyword, replace void by blank ([@vonglasow])
 * [#643](https://github.com/atoum/atoum/pull/643) atoum now requires PHP `>=5.6.0` ([@jubianchi])
 * [#634](https://github.com/atoum/atoum/pull/634) Only one extension of a kind can be loaded. Extensions can be unloaded ([@agallou], [@jubianchi])
 * [#619](https://github.com/atoum/atoum/pull/619) Add branches and paths coverage support to AtoumTask for Phing ([@oxman])
--- a/classes/report/fields/runner/tests/blank.php
+++ b/classes/report/fields/runner/tests/blank.php
@@ -8,7 +8,7 @@ use
 	mageekguy\atoum\observable
 ;
 
-abstract class void extends report\field
+abstract class blank extends report\field
 {
 	protected $runner = null;
 
--- a/classes/report/fields/runner/tests/blank/cli.php
+++ b/classes/report/fields/runner/tests/blank/cli.php
@@ -1,6 +1,6 @@
 <?php
 
-namespace mageekguy\atoum\report\fields\runner\tests\void;
+namespace mageekguy\atoum\report\fields\runner\tests\blank;
 
 use
 	mageekguy\atoum,
@@ -9,7 +9,7 @@ use
 	mageekguy\atoum\cli\colorizer
 ;
 
-class cli extends report\fields\runner\tests\void
+class cli extends report\fields\runner\tests\blank
 {
 	protected $titlePrompt = null;
 	protected $titleColorizer = null;
--- a/classes/reports/asynchronous/vim.php
+++ b/classes/reports/asynchronous/vim.php
@@ -95,7 +95,7 @@ class vim extends reports\asynchronous
 
 		$this->addField($uncompletedField);
 
-		$voidField = new runner\tests\void\cli();
+		$voidField = new runner\tests\blank\cli();
 		$voidField
 			->setTitlePrompt($firstLevelPrompt)
 			->setMethodPrompt($secondLevelPrompt)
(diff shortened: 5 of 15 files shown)


Grummfy:
yeahhh ;)

vonglasow:
@mikaelrandy The impact can be important if someone used the constant or even the class. It's for example the case on the project [hoaproject/Test](https://github.com/hoaproject/Test). But, we need to do it due to the change on PHP-7.1. On all extensions managed by atoum, I have seen only one require modification and the Pull Request is already done.

mikaelrandy:
@vonglasow hum ...  i'm still not confortable with the meaning of each class :/

But okay. It seems it's possible to use this class outside of the project itself, so it's a BC break and should lead to a new major release.

jubianchi:
@mikaelrandy we should confirm that it's not possible to alias the class from `blank` to `void` before planning a new major.

If we can add an alis in the atoum autoloader, we will not break anything ;)

vonglasow:
```sh
php -v
PHP 7.1.0alpha2 (cli) (built: Jun 27 2016 22:39:02) ( NTS )
Copyright (c) 1997-2016 The PHP Group
Zend Engine v3.1.0-dev, Copyright (c) 1998-2016 Zend Technologies
```
```Success (235 tests, 1870/1888 methods, 0 void method, 18 skipped methods, 27535 assertions)!```

**Comment by vonglasow** [comment]
@jubianchi I have added the alias in autoloader just tell me if it's ok for you.

**Comment by Hywan** [comment]
Are we good with this issue?

**Comment by vonglasow** [comment]
For me yes just need @jubianchi review. I will rebase it if it's ok before the merge

File: classes/autoloader.php [hunk 1]
```diff
@@ -41,7 +41,17 @@ public function __construct(array $namespaces = array(), array $namespaceAliases
...
 		}
 
-		foreach ($classAliases ?: array('atoum' => __NAMESPACE__ . '\test', __NAMESPACE__ => __NAMESPACE__ . '\test') as $alias => $target)
+		$defaultAliases = array(
+			'atoum' => __NAMESPACE__ . '\test',
+			__NAMESPACE__ => __NAMESPACE__ . '\test'
+		);
+
+		// Remove BC break for PHP_VERSION lower than 7.1.0 void is configured to be an alias of blank.
+		if (version_compare(PHP_VERSION, '7.1.0alpha2') < 0) {
```
jubianchi: I would have made this alias always available (i.e on all php versions)

vonglasow: Unfortunately it's not possible if I remove the condition it raise a fatal Error

```
Fatal error: Cannot use 'mageekguy\atoum\report\fields\runner\tests\void' as class name as it is reserved in /Users/ashgenesis/dev/atoum/classes/autoloader.php on line 244
```

jubianchi: my only fear here is when conditionally adding the alias we introduce some kind of BC break.

Imagine the following : I coded a custom report for atoum and I used the old `void` class. Everything will work fine on PHP < `7.1`. But when bumping my runtime, I will end up with an error. 

This can't be avoided because we can't `extends` a class conditionally so I'll have to publish two version of my report, one for PHP `< 7.1` and another for `>= 7.1`.

The only valid thing we can do here is just drop the old `void` class and rename it right away to `blank` but this is clearly a BC break in atoum and we must bump to `3.x`.

Bumping atoum means reviewing **every** published extension to adapt the version constraint in the `composer.json` which is a lot of work.

So the only valid patch is, to me, the one were we rename the old class and do not provide an alias. Then we will bump atoum to `3.0`. But I don't want to start a `3.0` without a tiny roadmap for this version.

Moreover, I think before bumping to 3.0, we should merge everything that is ready for 2.x so that our users will have the most up-to-date features in their projects without needing to bump to a new major.

What do you think @atoum/core @atoum/contributors

vonglasow: Completely agree, and I confirm this will introduce a BC Break. I agree with you the alias is not really useful here. And, I agree with you again the only patch valid is to rename the class void et don't provide any alias.

So do you need any help to valid or merge all PR ready for 2.x ?

jubianchi: @vonglasow if we merge this, it will be on a 3.0.

kao98:
I agree with you guys

Hywan:
Agree.

jdecool:
Do you know when the v3 will be release ?

jubianchi:
@jdecool I'll start working on the roadmap for the 3.x soon. I want to finish some WIP for the 2.9.0 then I'll be able to define what will be in the 3.x.

Once we know where we want to go, we'll start by merging this PR and tag a first release from the 3.x branch.

jubianchi:
@vonglasow can you enable the checkbox allowing me to work on your PR: I'd like to remove the alias in the autoloader so we can get this merged in the `3.0.0-dev` branch.

vonglasow:
@jubianchi done

shouze:
@jubianchi @Hywan @vonglasow any objection to backport this one into atoum 2.9? This would help folks that want to switch to (now released) php 7.1

I can make the PR if any of you don't have many time, but I'll need a maintainer for a 2.9.x tag.

Hywan:
It will introduce a BC break, so we keep it in 3.x serie.

shouze:
ho yes sorry...

Hywan:
Don't be sorry, it's fine :-).

shouze:
ok, I've read the roadmap to 3.0 and it's almost done, can't some non urgent thing could be postponed other updates?
Only split of some parts in other packages/repos seems to miss for a 3.0 release in fact.

Hywan:
Can you move this comment into #641? I guess this is more appropriate. Thanks!

omansour:
dear lord, some ppl in my team want to switch to phpunit in order to use 7.1 :(

KuiKui:
@omansour Pareil par ici.

Grummfy:
the beta version of atoum 3 include this one.

mikaelrandy:
@omansour @KuiKui does the [3.0.x-dev](https://packagist.org/packages/atoum/atoum#3.0.x-dev) branch a solution for you allowing to use PHP 7.1 ?

vdechenaux:
@mikaelrandy @KuiKui :+1: But he PHP version requirement is `^5.6.0 || ^7.0.0`

shouze:
@omansour @KuiKui @vdechenaux ok so plz use this ß guys!

Thx @mikaelrandy for releasing it, we're currently testing it on some projects, we will give you some feedback soon but as usual we expect atoum to make the job right™.

mikaelrandy:
@vdechenaux `^7.0.0` is supposed to mean `>=7.0.0, <8.0.0` (see [here](https://getcomposer.org/doc/articles/versions.md#caret))
Do you have any trouble to install atoum with this requirement ?

@shouze feedbacks will be wellcomed ;)

Issue #243, opened 2013-08-02
command line switches do not override .atoum.php configuration directives


gauthier:
I think that switches passed on CLI should take the priority over those in configuration file.

Relates To #242 (as a side-effect example of giving priority to .atoum.php file)

Hywan:
Is it still opened?

jubianchi:
@Hywan yes!

jubianchi:
* [x] `-f`/`--file` overrides `->files()` and `->directories()` from configuration file
* [x] `-d`/`--directories` overrides `->files()` and `->directories()` from configuration file
* [x] `-p`/`--php` overrides `->php()` from configuration file
* [x] `-mcn`/`--max-children-number` overrides `->maxChildrenNumber()` from configuration file
* [ ] `-ncc*`/`--no-code-coverage-*` should override configuration file
* [ ] `-tfe`/`--test-file-extensions` should override configuration file
* [ ] `-g`/`--glob` should override configuration file
* [ ] `-t`/`--tags` should override configuration file
* [ ] `-m`/`--methods` should override configuration file
* [ ] `-ns`/`--namespaces` should override configuration file
* [ ] `-af`/`--autoloader-file` should override configuration file
* [ ] `-bf`/`--bootstrap-file` should override configuration file
* [ ] `-ulr`/`--use-light-report` should override configuration file
* [ ] `-utr`/`--use-tap-report` should override configuration file
* [ ] `-xc`/`--xdebug-config` should override configuration file

Grummfy:
@jubianchi how can we help on that one? Can you point some line of code where we can start to look at ?

jubianchi:
@Grumfy The fix for this issue is available in #616.

I'll extract the code to fix this issue and postpone the verbose mode.

Grummfy:
thanks

PR #616, opened 2016-07-11
Add more information about the current configuration in verbose mode

### LINKED ISSUE #603: exporting configuration
hello,
to help to create plugins or to debug some strange case it could be usefull to have the possibility to export the configuration.

something like --export-config / -ec that will show the config

extra : 
having multiple configuration format available like json/init or simple human readable

jubianchi:
This will fix #603

While working on this, we will check if CLI arguments correctly override the configration files (see #243)

### DIFF
--- a/classes/cli/table.php
+++ b/classes/cli/table.php
@@ -1,0 +1,156 @@
+<?php
+
+namespace mageekguy\atoum\cli;
+
+use mageekguy\atoum\exceptions\runtime;
+
+class table
+{
+	private $rows = array();
+	private $rowMaxLength = 0;
+	private $columnsWidth = array();
+	private $header;
+
+	public function __construct(array $header = null)
+	{
+		$this->setHeader($header);
+	}
+
+	public function __toString()
+	{
+		$columnsWidth = $this->columnsWidth;
+		$table = '';
+
+		if ($this->header !== null)
+		{
+			$header = array_pad($this->header, $this->rowMaxLength, '');
+
+			$table .= '| ' . implode(' | ',
+				array_map(
+					function($index, $column) use ($columnsWidth) {
+						return str_pad($column, $columnsWidth[$index]);
+					},
+					array_keys($header),
+					array_values($header)
+				)
+			) . ' |' . PHP_EOL;
+
+			for ($i = 0; $i < $this->rowMaxLength; $i++)
+			{
+				$table .= '|' . str_repeat('-', $this->columnsWidth[$i] + 2);
+			}
+
+			$table .= '|' . PHP_EOL;
+		}
+
+		foreach ($this->rows as $row)
+		{
+			$row = array_pad($row, $this->rowMaxLength, '');
+
+			$table .= '| ' . implode(' | ',
+				array_map(
+					function($index, $column) use ($columnsWidth) {
+						return str_pad($column, $columnsWidth[$index]);
+					},
+					array_keys($row),
+					array_values($row)
+				)
+			) . ' |' . PHP_EOL;
+		}
+
+		$this->reset();
+
+		return $table . PHP_EOL;
+	}
+
+	public function setHeader(array $header = null)
+	{
+		if ($header === null)
+		{
+			return $this;
+		}
+
+		$this->header = array_values($header);
+
+		return $this->computeRowMaxLength($this->header)->computeColumnWidths($this->header);
+	}
+
+	public function addRow(array $row)
+	{
+		$row = array_values($row);
+
+		for ($index = 1, $size = sizeof($row); $index <= $size; $index++)
+		{
+			if (is_array($row[$index - 1]) === true)
+			{
+				if ($index < $size)
+				{
+					throw new runtime('Only the last column can be repeated');
+				}
+
+				if (sizeof($row[$index - 1]) === 1)
+				{
+					$row[$index - 1] = array_pop($row[$index - 1]);
+				}
+			}
+		}
+
+		if (is_array($row[sizeof($row) - 1]) === false)
+		{
+			$this->computeColumnWidths($this->rows[] = $row);
+		}
+		else
+		{
+			$lastColumn = array_pop($row);
+			$this->computeColumnWidths($this->rows[] = array_merge($row, array(array_shift($lastColumn))));
+
+			foreach ($lastColumn as $column)
+			{
+				$this->computeColumnWidths($this->rows[] = array_merge(array_fill(0, sizeof($row), ''), array($column)));
+			}
+		}
+
+		return $this->computeRowMaxLength($row);
+	}
+
+	public function reset()
+	{
+		$this->rows = array();
+		$this->rowMaxLength = 0;
+		$this->columnsWidth = array();
+
+		if ($this->header !== null)
+		{
+			$this->computeRowMaxLength($this->header)->computeColumnWidths($this->header);
+		}
+
+		return $this;
+	}
+
+	private function computeRowMaxLength(array $row)
+	{
+		$rowLength = sizeof($row);
+
+		if ($rowLength > $this->rowMaxLength)
+		{
+			$this->rowMaxLength = $rowLength;
+		}
+
+		return $this;
+	}
+
+	private function computeColumnWidths(array $row)
+	{
+		foreach ($row as $index => $column)
+		{
+			$columnWidth = strlen($column);
+
+			if (isset($this->columnsWidth[$index]) === false || $columnWidth > $this->columnsWidth[$index])
+			{
+				$this->columnsWidth[$index] = $columnWidth;
+			}
+		}
+
+		return $this;
+	}
+}
--- a/classes/iterators/filters/recursives/extension.php
+++ b/classes/iterators/filters/recursives/extension.php
@@ -36,11 +36,9 @@ class extension extends \recursiveFilterIterator
 
 	public function accept()
 	{
-		$path = basename((string) $this->getInnerIterator()->current());
+		$path = $this->getInnerIterator()->current();
 
-		$extension = pathinfo($path, PATHINFO_EXTENSION);
-
-		return ($extension == '' || in_array($extension, $this->acceptedExtensions) === true);
+		return ($path->isDir()|| in_array($path->getExtension(), $this->acceptedExtensions) === true);
 	}
 
 	public function getChildren()
--- a/classes/runner/visitors/verbose.php
+++ b/classes/runner/visitors/verbose.php
@@ -1,0 +1,61 @@
+<?php
+
+namespace mageekguy\atoum\runner\visitors;
+
+use
+	mageekguy\atoum\runner,
+	mageekguy\atoum\cli\table,
+	mageekguy\atoum\runner\visitor
+;
+
+class verbose implements visitor
+{
+	private $table;
+	private $verbosity;
+
+	public function __construct($verbosity = null, table $table = null)
+	{
+		$this->table = $table ?: new table(array('Runner property', 'value'));
+		$this->verbosity = (int) $verbosity;
+	}
+
+	public function __toString()
+	{
+		return (string) $this->table;
+	}
+
+	public function visit(runner $runner)
+	{
+		$this->table
+			->addRow(array('Runner class', get_class($runner)))
+			->addRow(array('Autoloader file', $runner->getAutoloaderFile() === null ? 'N/A' : $runner->getAutoloaderFile()))
+			->addRow(array('Bootstrap file', $runner->getBootstrapFile() === null ? 'N/A' : $runner->getBootstrapFile()))
+			->addRow(array('Max children processes', $runner->getMaxChildrenNumber() === null ? 'N/A' : $runner->getMaxChildrenNumber()))
+			->addRow(array('Reports', sizeof($runner->getReports()) === 0 ? 'N/A' : array_map(function($report) { return get_class($report); }, $runner->getReports())))
+			->addRow(array('xDebug configuration', $runner->getXdebugConfig() === null ? 'N/A' : $runner->getXdebugConfig()))
+			->addRow(array('Accepted file extensions', join(', ', $runner->getTestDirectoryIterator()->getAcceptedExtensions())))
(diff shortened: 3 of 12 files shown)


jubianchi:
We now have two things:

* CLI arguments correctly override configuration file (when possible)
  * `-g`/`--glob` cannot override the configuration file because its effects are immediatly applied to the runner
  * `-af`/`--autoloader-file` cannot override the configuration file because its effects are immediatly applied to the runner
  * `-bf`/`--bootstrap-file` cannot override the configuration file because its effects are immediatly applied to the runner
* Verbose mode displays a lot more informations:

```
Using ++verbose -g ./test*.php -ncc -xc IDEKEY=foobar arguments…
Using '/Users/julien.bianchi/repositories/atoum/atoum/.atoum.php' configuration file…
Using '/tmp/6e2deb60e60a380bb8dbe7b34cd86960.atoum.cache' autoloader cache file…
Using 'mageekguy\atoum\reports\realtime\santa' report…
Using 'IDEKEY=foobar' xDebug configuration…
Accepting 'php' file extensions…
Using './test.php' test file…
Using './test2.php' test file…
Code coverage is disabled…
```

any reviewer ?

Hywan:
Looking.

classes/scripts/runner.php
```diff
@@ -451,7 +493,7 @@ public function addTestsFromDirectories(array $directories)
 
 	public function addTestsFromPattern($pattern)
 	{
-		$this->runner->addTestsFromPattern($pattern);
+			$this->runner->addTestsFromPattern($pattern);
```
Hywan: Not sure about this one.

classes/scripts/runner.php
```diff
@@ -1208,11 +1264,87 @@ protected function doRun()
...
+				{
+					$this->verbose($this->locale->_('Running \'%s::%s\' test method…', $testClass, $testMethod), 3);
+				}
+			}
+
+			$maxChildrenNumber = $this->runner->getMaxChildrenNumber();
+
+			if ($maxChildrenNumber !== null)
+			{
+				$this->verbose(sprintf($this->locale->__('Limitintg children number to \'%d\' process…', 'Limitintg children number to \'%d\' processes…', $maxChildrenNumber), $maxChildrenNumber), 2);
```
Hywan: Limiting*

resources/configurations/runner/santa.php.dist
```diff
@@ -18,4 +18,4 @@ Generate a Santa report.
 $santaReport = new atoum\reports\realtime\santa();
 $santaReport->addWriter($stdOutWriter);
 
-$runner->addReport($cliReport);
+$runner->addReport($santaReport);
```
Hywan: lulz

Hywan:
@jubianchi: The workd “information“ in English is “uncountable”, there is no plural form. So it's many and several of hundreds of information (no “s”).

Hywan:
Instead of having `Using …` it would be better to reverse the display:
```
XDebug: …
Autoloader: …
```
Better than sentence I guess.

Grummfy:
thanks. But for the export of configuration I was also thinking extracting if we enable options or set stuff like : loop, debug, path to test directory, namespace of test,... the idea back #603 is to parse the solution, and get information on atoum for the IDE, or the debuging.

Honestly I can't review the code because I'm not advanced enough in the core of atoum. sorry :(

jubianchi:
@Hywan I edited the code so the output looks like this:

```
CLI arguments: ++verbose -g ./test*.php -xc IDEKEY=foobar -utr -ulr -ns foo bar\foo -mcn 3 -ebpc -nccfns foo -nccfm foo::bar -nccfc foo\bar\baz
Configuration file:
=> /Users/julien.bianchi/repositories/atoum/atoum/.atoum.php
Using no autoloader file…
Using no bootstrap file…
Autoloader cache file:
=> /var/folders/4g/16c91c3d507_645p1zszf4rc0000gp/T/6e2deb60e60a380bb8dbe7b34cd86960.atoum.cache
Reports:
=> mageekguy\atoum\reports\realtime\tap
xDebug configuration: IDEKEY=foobar
Accepted file extension: php
Test files:
=> ./test.php
=> ./test2.php
Accepted namespaces:
=> foo
=> bar\foo
Max children processes: 3
Code coverage:
=> Lines coverage
=> Branches coverage
=> Paths coverage
=> Excluded namespace:
==> foo
=> Excluded class:
==> foo\bar\baz
=> Excluded method:
==> foo::bar
```

@Grummfy I understand. This is the first step on something to quickly expose the configuration. We'll see if we can do anything better later, because, honestly, I don't think creating an exportable file would help ore than a clear output that user can copy/paste anywhere.

Hywan:
@jubianchi: If you have one autoload cache, why the newline and the `=>`? Same for reports.

jubianchi:
@Hywan In my configuration, I have only one autoloader and one report but we can add more ;)

jubianchi:
Here is the final output for the verbose mode:

| Property                 | Value                                                                                                                      |
|--------------------------|----------------------------------------------------------------------------------------------------------------------------|
| Script class             | mageekguy\atoum\scripts\coverage                                                                                           |
| CLI arguments            | -f /Users/julien.bianchi/repositories/atoum/atoum-config-extension/tests/units/classes/extension.php +verbose -o /dev/null |
| Autoloader cache file    | /var/folders/4g/16c91c3d507_645p1zszf4rc0000gp/T/6e2deb60e60a380bb8dbe7b34cd86960.atoum.cache                              |
| Configuration file       | N/A                                                                                                                        |
| Accepted tags            | N/A                                                                                                                        |
| Accepted namespaces      | N/A                                                                                                                        |
| Accepted methods         |                                                                                                                            |
| Runner class             | mageekguy\atoum\runner                                                                                                     |
| Autoloader file          | N/A                                                                                                                        |
| Bootstrap file           | N/A                                                                                                                        |
| Max children processes   | N/A                                                                                                                        |
| Reports                  | mageekguy\atoum\reports\realtime\cli                                                                                       |
|                          | mageekguy\atoum\reports\asynchronous\clover                                                                                |
| xDebug configuration     | N/A                                                                                                                        |
| Accepted file extensions | php                                                                                                                        |
| Test files               | 1                                                                                                                          |
| Code coverage            | Lines coverage                                                                                                             |
| CC excluded namespaces   | N/A                                                                                                                        |
| CC excluded classes      | N/A                                                                                                                        |
| CC excluded methods      | N/A                                                                                                                        |

As you can see, the output is a perfectly valid markdown table. Users will be able to copy/paster this directly in issues when they have problems.

Grummfy:
I test it, but is there a way to print the configuration but without executing the test?

Grummfy:
and one of the missing information is the namespace mask used for the unit test

jubianchi:
@Grummfy the namespace mask is not really something we can configure: we can't change it through a configuration file or using a CLI option.

Printing this information is hard right now as we would have to create an instance of the base test class and this class is abstract.

We can't extract this using reflection either because we would have to look for a constant, an annotation somewhere, a method implementation, ... too much things to lookup.

Before showing this we would have to refactor how atoum finds and uses this namespace pattern.

classes/iterators/filters/recursives/extension.php
```diff
@@ -36,11 +36,9 @@ public function getAcceptedExtensions()
 
 	public function accept()
 	{
```
Grummfy: change have no real link to the PR, but OK

classes/script/visitor.php
```diff
@@ -0,0 +1,19 @@
+<?php
+
+namespace mageekguy\atoum\script;
+
+use
+	mageekguy\atoum\script,
+	mageekguy\atoum\scripts
+;
+
+interface visitor
```
Grummfy: should we extends the previous existing visitor interface?

classes/scripts/runner.php
```diff
@@ -809,10 +861,13 @@ function($script, $argument, $empty) {
 					function($script, $argument, $directories) {
 						if (sizeof($directories) <= 0)
 						{
-							throw new exceptions\logic\invalidArgument(sprintf($script->getLocale()->_('Bad usage of %s, do php %s --help for more informations'), $argument, $script->getName()));
+							throw new exceptions\logic\invalidArgument(sprintf($script->getLocale()->_('Bad usage of %s, do php %s --help for more information'), $argument, $script->getName()));
 						}
 
-						$script->excludeDirectoriesFromCoverage($directories);
+						$script
+							->resetExcludedDirectoriesFromCoverage()
```
Grummfy: je comprends pas trop le rapport avec la PR

classes/scripts/runner.php
```diff
@@ -1170,48 +1239,14 @@ protected function doRun()
...
+			$table = new cli\table(array('Property', 'Value'));
+			$scriptVisitor = new atoum\script\visitors\verbose($table);
+			$runnerVisitor = new atoum\runner\visitors\verbose($this->getVerbosityLevel(), $table);
 
-			if ($autoloaderFile !== null)
-			{
-				$this->verbose($this->locale->_('Using \'%s\' autoloader file…', $autoloaderFile));
-			}
+			$scriptVisitor->visitRunner($this);
+			$runnerVisitor->visit($this->getRunner());
```
Grummfy: nothing for the extensions?

scripts/coverage.php
```diff
@@ -1,34 +1,17 @@
 <?php
```
Grummfy: don't see why this change

tests/units/classes/iterators/filters/recursives/extension.php
```diff
@@ -40,16 +40,41 @@ public function test__construct()
 	public function testAccept()
```
Grummfy: idem, je voix pas trop le rapport ...

par contre pas de unit test pour la sortie?

classes/cli/table.php
```diff
@@ -0,0 +1,156 @@
...
+		$this->header = array_values($header);
+
+		return $this->computeRowMaxLength($this->header)->computeColumnWidths($this->header);
+	}
+
+	public function addRow(array $row)
+	{
+		$row = array_values($row);
+
+		for ($index = 1, $size = sizeof($row); $index <= $size; $index++)
```
Grummfy: we should start $index at 0, and "$index < $size" should become "$index < ($size - 1)", no?

classes/cli/table.php
```diff
@@ -0,0 +1,156 @@
...
+				}
+
+				if (sizeof($row[$index - 1]) === 1)
+				{
+					$row[$index - 1] = array_pop($row[$index - 1]);
+				}
+			}
+		}
+
+		if (is_array($row[sizeof($row) - 1]) === false)
```
Grummfy: sizeof($row) => $size

classes/cli/table.php
```diff
@@ -0,0 +1,156 @@
...
+			$lastColumn = array_pop($row);
+			$this->computeColumnWidths($this->rows[] = array_merge($row, array(array_shift($lastColumn))));
+
+			foreach ($lastColumn as $column)
+			{
+				$this->computeColumnWidths($this->rows[] = array_merge(array_fill(0, sizeof($row), ''), array($column)));
+			}
+		}
+
+		return $this->computeRowMaxLength($row);
```
Grummfy: honestly I will do all this merge and row mix a the end, no?

classes/scripts/runner.php
```diff
@@ -1170,48 +1239,14 @@ protected function doRun()
...
-			if (sizeof($this->configFiles) > 0)
-			{
-				foreach ($this->configFiles as $configFile)
-				{
-					$this->verbose($this->locale->_('Using \'%s\' configuration file…', $configFile));
-				}
-			}
-
-			$autoloaderFile = $this->runner->getAutoloaderFile();
+			$table = new cli\table(array('Property', 'Value'));
```
Grummfy: why not a big if is verbose? or following if it's verbose or not, the visitor do something or nothing (dev/null mode)

Grummfy: use $this->verbosityLevel ?

Grummfy: perhaps we can have an interface for that part, with no default implementation (why a default one ?)

and then like the rest create a setVerboseRenderer(verboseRendererInterface $vr)? with a default renderer == table ?

classes/runner/visitors/verbose.php
```diff
@@ -0,0 +1,61 @@
...
+
+	public function __toString()
+	{
+		return (string) $this->table;
+	}
+
+	public function visit(runner $runner)
+	{
+		$this->table
+			->addRow(array('Runner class', get_class($runner)))
```
agallou: table utilisé dans le visiteur. ça ne lie pas le visiteur avec la vue ? ne faudrait t-il pas passer par un objet contenant ces infos ?

classes/script/visitor.php
jubianchi: which one ?

classes/scripts/runner.php
jubianchi: see #243

tests/units/classes/iterators/filters/recursives/extension.php
jubianchi: see #243

classes/scripts/runner.php
```diff
@@ -1170,48 +1239,14 @@ protected function doRun()
...
+			$scriptVisitor = new atoum\script\visitors\verbose($table);
+			$runnerVisitor = new atoum\runner\visitors\verbose($this->getVerbosityLevel(), $table);
```
agallou: utiliser le même constructeur pour les visitors ? (qui serait dans l'interface)

classes/script/visitor.php
Grummfy: the other visitor class , but in fact no? just same name but differnt purpose ;)

classes/cli/table.php
jubianchi: if we do this just before the table rendering we have to do one more complete iteration over the whole table. Doing it here lets us compute values each time a row is added (so no global iteration on the whole table) and have those values directly available when rendering.

jubianchi:
@Grummfy I'm not satisfied with the current implementation. It's poorly designed.

I'd like to close this issue so that we can start on something cleaner.

PR #618, opened 2016-07-12
Fix wrong detection of reports_extension

oxman:


resources/phing/AtoumTask.php
```diff
@@ -180,7 +180,7 @@ public function execute()
 
 		if ($this->telemetryEnabled())
 		{
-			if (class_exists('reports\telemetry') === false) {
+			if (class_exists('mageekguy\atoum\reports\telemetry') === false) {
```
mikaelrandy: Is it possible to have any side effect with "magical" `mageekguy\atoum` to `atoum` translation ?

oxman: I don't think

PR #625, opened 2016-08-15
Various HHVM compatibilities

### LINKED ISSUE #624: HHVM incompatibility with `directory\factory::getIterator`
See https://github.com/atoum/atoum/pull/479#issuecomment-239820823.
Long story short: The `mageekguy\atoum\iterators\recursives\directory\factory` class is implementing `IteratorAggregate`, but the `getIterator` method expects an argument, which is wrong.

We need to fix it.

Hywan:
Fix #624.

Various stuff here.

First, HHVM detects an incorrect interface implementation. Don't know why Zend Engine doesn't… So the `IteratorAggregate` interface is removed from the `directory\factory` class.

Second, HHVM binary requires the `--php` to run a PHP file. This is required for the isolation test execution engine.

Third, HHVM nightly is no longer build (see https://github.com/travis-ci/travis-ci/issues/3788) so we can remove it from Travis configuration file.

### DIFF
--- a/classes/php.php
+++ b/classes/php.php
@@ -41,6 +41,10 @@ class php extends cli\command
 			}
 		}
 
+		if (defined('HHVM_VERSION')) {
+			$phpPath .= ' --php';
+		}
+
 		return parent::setBinaryPath($phpPath);
 	}
 }
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,9 +25,12 @@ before_script:
   - "[ -n \"$COVERALLS_REPO_TOKEN\" ] && sed -i s/YouCoverallsProjectToken/$COVERALLS_REPO_TOKEN/ .atoum.php || true"
 script:
   - bin/atoum +verbose
-  - mkdir ../phar
-  - php -n -dphar.readonly=Off scripts/phar/generator.php -d ../phar
-  - php -n -ddate.timezone=Europe/Paris ../phar/atoum.phar --test-it -ncc +verbose -c .atoum.php
+  - |
+      if [[ $TRAVIS_PHP_VERSION = php* ]]; then
+        mkdir ../phar
+        php -n -dphar.readonly=Off scripts/phar/generator.php -d ../phar
+        php -n -ddate.timezone=Europe/Paris ../phar/atoum.phar --test-it -ncc +verbose -c .atoum.php
+      fi
 deploy:
   provider: releases
   api_key:
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,10 +1,24 @@
 language: php
 matrix:
   include:
+    - php: 5.3
+    - php: 5.4
+    - php: 5.5
+    - php: 5.6
+    - php: 7.0
+    - php: nightly
     - php: hhvm-3.12
       sudo: required
       dist: trusty
       group: edge
+    - php: hhvm
+      sudo: required
+      dist: trusty
+      group: edge
+  allow_failures:
+    - php: nightly
+    - php: hhvm-3.12
+    - php: hhvm
   fast_finish: true
 notifications:
   irc: "irc.freenode.org##atoum"
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,15 +1,10 @@
 language: php
-php:
-  - 5.3
-  - 5.4
-  - 5.5
-  - 5.6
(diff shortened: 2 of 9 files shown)


mikaelrandy:
I ask myself why we originally set `directory\factory` as an `IteratorAggregate` and what this change could broke.
Maybe it's a BC break update ?

Hywan:
@mikaelrandy I am not sure this is a BC break because it was wrong. I don't even understand why Zend Engine does not throw an error at this… For me this is a bug fix.

Hywan:
This PR contains another patch for HHVM compatibility. I am not able to run atoum on HHVM without these patches.

Hywan:
I have updated my PR description message to gather all the information.

Hywan:
So… after a lot of patches, here is the state of this PR.

1. We can **use** atoum to run our tests on HHVM (this is new, the concurrent engine was not working),
2. atoum can run its own test suites on HHVM (this is new, various issues were blocking this),
3. atoum test suites fails on HHVM (this is new, we discover this result).

Hywan:
Also, now we are testing against HHVM 3.12 (an LTS) and HHVM 3.14 (the latest). Both are allowed failures.

Hywan:
Ready for a review @mikaelrandy!

Hywan:
HHVM is very very slow…

jubianchi:
before merging this, we have to be sure that

* using ˋhhvm bin/atoum --php 'php -n'ˋ works
* using ˋphp bin/atoum --php 'hhvm --php' works

this is an important thing to consider: the runner and the child processes can use a different PHP executable.

Hywan:
@jubianchi This means executing the whole test suite 4 times:
  1. `php bin/atoum` (ZE with ZE),
  2. `hhvm bin/atoum` (HHVM with HHVM),
  3. `php bin/atoum --php 'hhvm --php' (ZE with HHVM),
  4. `hhvm bin/atoum --php 'php -n' (HHVM with ZE).

I am not sure we can do that with Travis because it mixes PHP VM and we are testing against one specific VM at a time. I would be likely to try on my machine but HHVM no longer compile. Do you have HHVM?

Hywan:
@jubianchi Definitively HHVM is not building on my OS. I could still install Debian to test but… sigh.

While I understand your requirements, how to fullfill it? Do you want an automatic script on Travis (it's going to be difficult because we cannot have 2 VM installed at the same time easily), or just a manual test?

jubianchi:
@Hywan first, we need to make sure it works with manual tests. Then we will be able to test the behavior using the unit test suite.

Hywan:
@jubianchi What do you mean? It works on Travis.

jubianchi:
[This](https://github.com/atoum/atoum/pull/625#issuecomment-240660372) is not tested on travis. We have to manually test it.

Then, using unit tests, we will have to simulate this and make sure it's OK.

Hywan:
OK to test it manually if we can have a working HHVM. However, this is hard to create real unit test for it because we need both VM running side-by-side, and this is not possible on Travis as far as I know. Am I wrong?

Hywan:
ping @jubianchi 
I am kind of lost here. Could you help me for the next step?

classes/php.php
```diff
@@ -41,6 +41,10 @@ public function setBinaryPath($phpPath = null)
 			}
 		}
 
+		if (defined('HHVM_VERSION')) {
```
Hywan: I might want to move this here, https://github.com/atoum/atoum/blob/master/classes/test/engines/concurrent.php#L178-L181.

Grummfy:
@Hywan you need an upgrade from the master ;)
@jubianchi Hywan ask a little help on that one

Hywan:
HHVM is not compatible with PHP 7. Almost all projects are dropping its support. I reckon we should do the same. Thoughts?

mikaelrandy:
👍

Grummfy:
so we close?

Hywan:
@Grummfy Not before having more feedbacks about the dropping of HHVM :-).

jubianchi:
I think we can close all the issues and PRs related to HHVM.

We never had many requests about this and the commits will be easy to find if we ever have to apply them.

Hywan:
I agree :+1:!

PR #631, opened 2016-09-26
Coveralls report: do not send the repository token if it's not available

### LINKED ISSUE #630: Coveralls for all
Hello :-),

I am deploying Coveralls with atoum and I have an issue with it. It cannot work with forks because of Travis policy with environment variables. However, I recently used [`travis-cargo`](https://github.com/huonw/travis-cargo/) and it supports coveralls natively without any token on any forks. I was wondering if we can mimic what happens there to provide the same user experience: No need to set up a token on Travis, no need to set up a token from Coveralls, it works with the origin repository, it works with all the forks etc.

By reading the code, it means it calls the API of coveralls directly (from kcov), like this, https://github.com/SimonKagstrom/kcov/blob/master/src/writers/coveralls-writer.cc. It looks somewhat similar to https://github.com/atoum/atoum/blob/master/classes/reports/asynchronous/coveralls.php.

Did I miss something particular or can we provide the same user experience?

jubianchi:
Closes #630


Hywan:
:+1:

Hywan:
@jubianchi No changelog update?

PR #633, opened 2016-09-26
Mock generator correctly handles the `self` and `parent` return types

### LINKED ISSUE #623: Cannot mock class with typed return self
```
=> Skeleton\tests\units\Skeleton::testWithMock():
==> Error FATAL ERROR in /var/www/dev/atoum-skeleton/tests/units/Skeleton.php on unknown line, generated by file /var/www/dev/atoum-skeleton/vendor/atoum/atoum/classes/mock/generator.php(199) : eval()'d code on line 52:
'\self' is an invalid class name
> There is 1 uncompleted method:
=> Skeleton\tests\units\Skeleton::testWithMock() with exit code 255:
==> output(0) ""
Code coverage: 0.00%.
```
This above error occur when we tried to mock a class which typed their return
Example 
```
    public function fluent() : self
    {
        return $this;      
    }
```

A MWE could be found in this [repository](https://github.com/vonglasow/atoum-skeleton/tree/bug-self-mock)

https://wiki.php.net/rfc/return_types

>    Note that it does allow self and parent to be used as return types.

```
PHP 7.0.9-1+deb.sury.org~trusty+1 (cli) ( NTS )
Copyright (c) 1997-2016 The PHP Group
Zend Engine v3.0.0, Copyright (c) 1998-2016 Zend Technologies
    with Zend OPcache v7.0.9-1+deb.sury.org~trusty+1, Copyright (c) 1999-2016, by Zend Technologies
```

jubianchi:
Closes #623

```php
<?php

declare(strict_types=1);

namespace
{
    class foo
    {
        public function bar() : self
        {
            return $this;
        }
    }
}

namespace tests\units
{
    use atoum;

    class foo extends atoum
    {
        public function testBar()
        {
            $this->object(new \mock\foo())->isInstanceOf($this->newTestedInstance);
        }
    }
}
```

This works  without error:

```shell
bin/atoum mock-self.php 
> atoum path: /Users/julien.bianchi/repositories/atoum/atoum/bin/atoum
> atoum version: dev-master
> PHP path: /usr/local/Cellar/php70/7.0.9/bin/php
> PHP version:
=> PHP 7.0.9 (cli) (built: Jul 27 2016 15:51:40) ( ZTS )
=> Copyright (c) 1997-2016 The PHP Group
=> Zend Engine v3.0.0, Copyright (c) 1998-2016 Zend Technologies
=>     with Xdebug v2.4.1-dev, Copyright (c) 2002-2016, by Derick Rethans
=>     with blackfire v1.11.1, https://blackfire.io, by Blackfireio Inc.
> tests\units\foo...
[S___________________________________________________________][1/1]
=> Test duration: 0.01 second.
=> Memory usage: 0.00 Mb.
> Total test duration: 0.01 second.
> Total test memory usage: 0.00 Mb.
> Code coverage value: 0.00%
=> Class foo: 0.00%
==> foo::bar(): 0.00%
> Running duration: 0.07 second.
Success (1 test, 1/1 method, 0 void method, 0 skipped method, 2 assertions)!
```

### DIFF
--- a/classes/mock/generator.php
+++ b/classes/mock/generator.php
@@ -612,13 +612,22 @@ class generator
 
 		if($method->getName() !== '__construct' && method_exists($method, 'hasReturnType') && $method->hasReturnType())
 		{
-			if($method->getReturnType()->isBuiltin())
+			switch (true)
 			{
-				$returnTypeCode = ': ' . $method->getReturnType();
-			}
-			else
-			{
-				$returnTypeCode = ': \\' . $method->getReturnType();
+				case (string) $method->getReturnType() === 'self':
+					$returnTypeCode = ': \\' . $method->getDeclaringClass()->getName();
+					break;
+
+				case (string) $method->getReturnType() === 'parent':
+					$returnTypeCode = ': \\' . $method->getDeclaringClass()->getParentClass()->getName();
+					break;
+
+				case $method->getReturnType()->isBuiltin():
+					$returnTypeCode = ': ' . $method->getReturnType();
+					break;
+
+				default:
+					$returnTypeCode = ': \\' . $method->getReturnType();
 			}
 		}
 
--- a/tests/units/classes/mock/generator.php
+++ b/tests/units/classes/mock/generator.php
@@ -3447,6 +3447,90 @@ class generator extends atoum\test
 		;
 	}
 
+	/** @php >= 7.0 */
+	public function testGetMockedClassCodeForMethodWithSelfReturnType()
+	{
+		$this
+			->if($generator = new testedClass())
+			->and($reflectionTypeController = new mock\controller())
+			->and($reflectionTypeController->__construct = function() {})
+			->and($reflectionTypeController->__toString = 'self')
+			->and($reflectionTypeController->isBuiltIn = false)
+			->and($reflectionType = new \mock\reflectionType())
+			->and($reflectionMethodController = new mock\controller())
+			->and($reflectionMethodController->__construct = function() {})
+			->and($reflectionMethodController->getName = $methodName = 'returnSelf')
+			->and($reflectionMethodController->isConstructor = false)
+			->and($reflectionMethodController->getParameters = array())
+			->and($reflectionMethodController->isPublic = true)
+			->and($reflectionMethodController->isProtected = false)
+			->and($reflectionMethodController->isPrivate = false)
+			->and($reflectionMethodController->isFinal = false)
+			->and($reflectionMethodController->isStatic = false)
+			->and($reflectionMethodController->isAbstract = false)
+			->and($reflectionMethodController->returnsReference = false)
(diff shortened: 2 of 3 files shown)


classes/mock/generator.php
```diff
@@ -618,7 +618,7 @@ protected function getReturnType(\reflectionMethod $method)
 			}
 			else
 			{
-				$returnTypeCode = ': \\' . $method->getReturnType();
+				$returnTypeCode = ': \\' . ((string) $method->getReturnType() === 'self' ? $method->getDeclaringClass()->getName() : $method->getReturnType());
```
Hywan: :+1:

tests/units/classes/mock/generator.php
```diff
@@ -3447,6 +3447,90 @@ public function testGetMockedClassCodeForMethodWithReservedWord()
 		;
 	}
 
+    /** @php >= 7.0 */
+    public function testGetMockedClassCodeForMethodWithSelfReturnType()
+    {
+        $this
+			->if($generator = new testedClass())
```
Hywan: CS.

vonglasow:
It's ok for me :+1:

mageekguy:
For the record, make test on mock is… uhm, i don't know what it is exactly, but i'm pretty sure that it's a very bad idea.

jubianchi:
@mageekguy we are not talking about mocking the tested class, yes my example is bad. It's just a MWE ;)

The real problem is: the mock generator does not handle classes that return `self`. I had to patch this because we have to be able to mock anything.

But yeah, you are totally right: mocking the tested class is not good.

jubianchi:
I just pushed an update to this PR. The generator now also handles the `parent` return type:

```php
<?php

declare(strict_types=1);

namespace
{
    class foo
    {
        public function one() : self
        {
            return $this;
        }

        public function two() : self
        {
            return $this;
        }
    }

    class bar extends foo
    {
        public function one() : parent
        {
            return $this;
        }

        public function two() : parent
        {
            return $this;
        }

        public function three() : self
        {
            return $this;
        }
    }

    class baz extends bar
    {
        public function two() : foo
        {
            return $this;
        }

        public function three() : parent
        {
            return $this;
        }
    }
}

namespace tests\units
{
    use atoum;

    class stdClass extends atoum
    {
        public function testBar()
        {
            $this->object(new \mock\foo())->isInstanceOf(\foo::class);
            $this->object(new \mock\bar())->isInstanceOf(\foo::class);
            $this->object(new \mock\baz())
                ->isInstanceOf(\foo::class)
                ->isInstanceOf(\bar::class)
            ;
        }
    }
}
```

vonglasow:
It's ok for me all tests passed with this fix.

jubianchi:
@Hywan can you please check your review and approve the changes if they are ok.

PR #634, opened 2016-09-26
Only one extension of a kind can be loaded

jubianchi:
This is a partial port of what @agallou did in #604: extensions
are indexed by class name.

* We can load an extension the way we already did:

```php
<?php

$extension = new mageekguy\atoum\blackfire\extension($script);
$extension
    ->setClientConfiguration(new \Blackfire\ClientConfiguration(
        $_ENV['BLACKFIRE_CLIENT_ID'],
        $_ENV['BLACKFIRE_CLIENT_TOKEN']
    ))
    ->addToRunner($runner)
;
```

* If we later try to re-register the same extension, only its
  configuration will be updated:

```php
<?php

$extension = new mageekguy\atoum\blackfire\extension($script);
$extension
    ->setClientConfiguration(new \Blackfire\ClientConfiguration(
        'foo',
        'bar'
    ))
    ->addToRunner($runner)
;
```

* We can also unload an extension, either using an instance or its
  class name:

```php
$extension = new mageekguy\atoum\blackfire\extension($script);
$extension->addToRunner($runner);

$runner->removeExtension($extension);
$runner->removeExtension(new mageekguy\atoum\blackfire\extension());
$runner->removeExtension(mageekguy\atoum\blackfire\extension::class);
```

### DIFF
--- a/classes/extension/aggregator.php
+++ b/classes/extension/aggregator.php
@@ -1,0 +1,18 @@
+<?php
+
+namespace mageekguy\atoum\extension;
+
+use mageekguy\atoum\exceptions\logic\invalidArgument;
+
+class aggregator extends \splObjectStorage
+{
+    public function getHash($object)
+    {
+        if (is_object($object) === false)
+        {
+            throw new invalidArgument(__METHOD__ . ' expects parameter 1 to be object, ' . gettype($object) . ' given');
+        }
+
+        return get_class($object);
+    }
+}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 # `dev-master`
 
+* [#634](https://github.com/atoum/atoum/pull/634) Only one extension of a kind can be loaded. Extensions can be unloaded ([@agallou], [@jubianchi])
 * [#619](https://github.com/atoum/atoum/pull/619) Add branches and paths coverage support to AtoumTask for Phing ([@oxman])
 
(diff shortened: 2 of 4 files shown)


classes/extension/aggregator.php
```diff
@@ -0,0 +1,19 @@
+<?php
+
+namespace mageekguy\atoum\extension;
+
+
```
Hywan: One useless newline.

classes/extension/aggregator.php
```diff
@@ -0,0 +1,19 @@
...
+use mageekguy\atoum\exceptions\logic\invalidArgument;
+
+class aggregator extends \splObjectStorage
+{
+    public function getHash($object)
+    {
+        if (is_object($object) === false)
+        {
+            throw new invalidArgument(__METHOD__ . ' expects parameter 1 to be object, ' . gettype($object) . ' given');
```
Hywan: `getHash` already expect an object as its unique argument (http://php.net/manual/en/splobjectstorage.gethash.php). Is it really necessary to re-check it?

jubianchi: @Hywan we do not rely on the `splObjectStorage::getHash()` method here, we simply override it. We can't typehint on object so we check here.

Do you prefer a `stdClass` typehint ?

Hywan: No, `stdClass` is not the parent of any classes. Let's keep it like this!

PR #639, opened 2016-09-30
Update the license so that @mageekguy is not the only copyright holder

jubianchi:


LICENSE
```diff
@@ -1,3 +1,4 @@
+Copyright (c) 2016, atoum contributors.
```
Hywan: Would it not be better to have `2016-2016` for a better scripting update?

README.md
```diff
@@ -339,3 +339,7 @@ Looking for a roadmap?
 
 - [Here is](https://github.com/atoum/atoum/labels/In%20progress) the work in progress,
 - [And there](https://github.com/atoum/atoum/milestones/*) what will come in next releases.
+
+## License
+
+atoum is released under the BSD-3-Clause License. See the bundled [LICENSE](LICENSE) file for details.
```
Hywan: I guess that `[LICENSE]()` still works.

CREDITS.md
```diff
@@ -1,11 +1,12 @@
-#Main contributors
+# Main contributors
 
+* Frédéric Hardy (frédéric.hardy@atoum.org)
 * Julien Bianchi (julien.bianchi@atoum.org)
 * Ludovic Fleury (ludovic.fleury@atoum.org)
 * Gérald Croes (gerald.croes@atoum.org)
 * François Dussert (francois.dussert@atoum.org)
```
Hywan: Is this list still up-to-date?
Does it make sense to keep it?

README.md
jubianchi: Yes it should work in Github but to be sure it works with most markdown parsers, I prefer to keep this as is.

CREDITS.md
jubianchi: I'll propose something else ;)

README.md
```diff
@@ -339,3 +339,25 @@ Looking for a roadmap?
 
 - [Here is](https://github.com/atoum/atoum/labels/In%20progress) the work in progress,
 - [And there](https://github.com/atoum/atoum/milestones/*) what will come in next releases.
+
+## Credits
+
+atoum has been created by [Frédéric Hardy]. It is now led by a strong community of contributors. You can find them 
+in the [committer list](https://github.com/atoum/atoum/graphs/contributors) or in the 
+[Contributors team](https://github.com/orgs/atoum/teams/contributors).
```
Hywan: Is it voluntary to have a capital C to Contributors?

jubianchi: yes: the team name has a capital "C" ;)

LICENSE
```diff
@@ -1,3 +1,4 @@
+Copyright (c) 2016-2016, atoum contributors.
 Copyright (c) 2010-2016, Frédéric Hardy.
```
mikaelrandy: 2 lines of Copyright ?

jubianchi: @mikaelrandy yes. If we were to change the license we would have to ask all contributor if they are OK with that.

Moreover, removing the name of @mageekguy from the license does not seem good to me: he was the copyright owner. We just choose to change this from now on.

Issue #641, opened 2016-10-01
The road to 3.0


jubianchi:
Everything in the following list is under discussion, feel free to send feedback on anything. Additionally, If you see something missing, please tell me.
- [x] Require a higher PHP version (at least `>= 5.6`) (#643)
- [x] #646 - Require a higher xDebug version (`>= 2.3.0`) and add it to `suggest` in `composer.json` (#651)
- [x] #653 - Remove all code relying on `method_exists`
  - [x] in `mageekguy\atoum\factory\builder\closure::isVariadic()`
  - [x] in `mageekguy\atoum\mock\generator::getParameterType()` (line 699)
  - [x] in `mageekguy\atoum\mock\generator::isVariadic()`
  - [x] in `mageekguy\atoum\php\mocker\funktion::getParameterType()`
  - [x] in `mageekguy\atoum\score\coverage::getDeclaringClass()` 
- [x] #647 - Remove all tests that are not relevant on the minimum PHP version (`< 5.6`) (#657)
- [x] #647 - Remove PHP version filter where it target `>= 5.6` (#657)
- [x] Remove all PHP versions that are not relevant from CI (#643)
- [x] #648 - Remove deprecated methods (#650)
  - [x] `mageekguy\atoum\asserters\dateTime::isInYear()`
  - [x] `mageekguy\atoum\asserters\dateTime::isInMonth()`
  - [x] `mageekguy\atoum\asserters\dateTime::isInDay()`
  - [x] `mageekguy\atoum\asserters\stream::isWrited()`
- [x] #652 - Remove deprecated CLI flags (#656)
  - [x] `--test-all` and its handler (in `mageekguy\atoum\scripts\runner`, line 1014)
  - [x] `mageekguy\atoum\scripts\runner::addTestAllDirectory()`
- [x] Merge #615 
- [x] Revert 9c6772ad6: closure are bound to te current object context, no need to pass the test instance anymore
- [x] Remove all `$this` assignation that are only for closures `use`s (#658)
- [x] #645 - Remove all code relying on `version_compare` (#649):
  - [x] in `mageekguy\atoum\autoloader::exists()` (`5.4.0`)
  - [x] in `mageekguy\atoum\asserters\dateTime::isImmutable()` (`5.5.0`)
  - [x] in `mageekguy\atoum\asserters\dateTime::isDateTime()` (`5.5.0`)
  - [x] in `mageekguy\atoum\test\adapter\invoker::isBindable()` (`5.4.0`)
[...]

PR #644, opened 2016-10-08
remove nyancat and santa reports

agallou:
(they has been moved to the reports extension)

cf #641


jubianchi:
closed with de4dc3d3704

Issue #653, opened 2016-10-08
Remove all code relying on `method_exists`


jubianchi:
* [x] in `mageekguy\atoum\factory\builder\closure::isVariadic()`
* [x] in `mageekguy\atoum\mock\generator::getParameterType()` (line 699)
* [x] in `mageekguy\atoum\mock\generator::isVariadic()`
* [x] in `mageekguy\atoum\php\mocker\funktion::getParameterType()`
* [x] in `mageekguy\atoum\score\coverage::getDeclaringClass()`

andyps:
Hi! Please, check pull request [#655 ]( https://github.com/atoum/atoum/pull/655)

PR #655, opened 2016-10-08
Remove all code relying on `method_exists`

andyps:
For [#653](https://github.com/atoum/atoum/issues/653)


jubianchi:
thanks @andyps!

Hywan:
Thanks!

PR #660, opened 2016-10-16
fix santa report

agallou:
spaces were replaced by strings in the printed string.

before:
capture du 2016-10-16 11 57 19

after:
capture du 2016-10-16 11 57 39

this fix does not have to also be on the 3.0 branch, but must be cherry-picked on this PR of the reports extension : https://github.com/atoum/reports-extension/pull/26

### DIFF
--- a/classes/report/fields/runner/event/santa.php
+++ b/classes/report/fields/runner/event/santa.php
@@ -32,24 +32,24 @@ class santa extends cli
 	$brown = new colorizer('38;5;94');
 
 	$this->sprite = array(
-	'			_			',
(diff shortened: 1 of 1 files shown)


Hywan:
@agallou Thanks!
@jubianchi I set the milestone to 3.0.0 but I reckon if 2.9.0 would not be better. Thoughts? cc @mikaelrandy

jubianchi:
@Hywan the PR is targeting the master branch so it will be in the 2.9.0. Moreover, we absolutely need this for sitepoint-editors/php-peers#369

PR #661, opened 2016-10-16
adds a generator asserter

agallou:
Example of usage:

```php
	public function testUsage()
	{
		$generator = function() {
			for ($i=0; $i<3; $i++) {
				yield ($i+1);
			}
		};

		$this
			->generator($generator())
				->yields->isEqualTo(1)
				->yields->isEqualTo(2)
				->yields->isEqualTo(3)
				->yields->isNull(4)
			->generator($generator())
				->size->isEqualTo(3)
		;
	}
```

all methods of the iterator asserter could be called on generator
the yields call returns a variable asserter on the yielded value.

see #641

### DIFF
--- a/tests/units/classes/asserters/generator.php
+++ b/tests/units/classes/asserters/generator.php
@@ -202,7 +202,7 @@ PHP
 			}
 		};
 
-		$notAgenerator = function() {
+		$notAGenerator = function() {
 			for ($i=0; $i<10; $i++) {
 			}
 		};
@@ -221,7 +221,7 @@ PHP
 					->hasMessage("boolean(true) is not an object")
 
 			->then
-				->exception(function() use ($asserter, $notAgenerator) { $asserter->setWith($notAgenerator()); })
+				->exception(function() use ($asserter, $notAGenerator) { $asserter->setWith($notAGenerator()); })
 					->isInstanceOf('mageekguy\atoum\asserter\exception')
 					->hasMessage("null is not an object")
 
--- a/tests/units/classes/asserters/generator.php
+++ b/tests/units/classes/asserters/generator.php
@@ -40,36 +40,6 @@ class generator extends atoum\test
 	}
 
 	/**
-	 * @php >= 7.0
-	 */
-	public function testUsage()
-	{
-		if (version_compare(PHP_VERSION, '7.0') >= 0) {
-			$generator = eval(<<<'PHP'
-return function() {
-    for ($i=0; $i<3; $i++) {
-        yield ($i+1);
-    }
-
-    return 42;
-};
-PHP
-			);
-		}
-
-		$this
-			->generator($generator())
-				->yields->isEqualTo(1)
-				->yields->isEqualTo(2)
-				->yields->isEqualTo(3)
-				->yields->isNull()
-				->returns->isEqualTo(42)
-			->generator($generator())
-				->size->isEqualTo(3)
-		;
-	}
-
-	/**
 	 * @php < 7.0
 	 */
 	public function testReturnsBeforePhp7()
--- a/tests/units/classes/asserters/generator.php
+++ b/tests/units/classes/asserters/generator.php
@@ -39,6 +39,25 @@ class generator extends atoum\test
 		;
 	}
 
+	public function testUsage()
+	{
+		$generator = function() {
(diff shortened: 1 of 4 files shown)


jubianchi:
@agallou awesome!

One thing is missing: as of PHP 7.0 generators can return a value so we have to provide an assertion for that.

```
public function testUsage()
{
    $generator = function() {
        for ($i=0; $i<3; $i++) {
            yield ($i+1);
        }

        return PHP_INT_MAX;
    };

    $this
        ->generator($generator())
            ->yields->isEqualTo(1)
            ->yields->isEqualTo(2)
            ->yields->isEqualTo(3)
            ->returns->isEqualTo(PHP_INT_MAX)
    ;
}
```

With both `yields` and `returns` assertions I think we are good for a first version of the asserter. We'll be able to make it better later (with things to handle `send`, `isValid`,  ...)

tests/units/classes/asserters/generator.php
```diff
@@ -0,0 +1,129 @@
...
+				yield ($i+1);
+			}
+		};
+
+		$this
+			->generator($generator())
+				->yields->isEqualTo(1)
+				->yields->isEqualTo(2)
+				->yields->isEqualTo(3)
+				->yields->isNull(4)
```
jubianchi: `isNull` should not take any argument, unless it is a fail message.

tests/units/classes/asserters/generator.php
```diff
@@ -0,0 +1,129 @@
...
+
+	public function testSetWith()
+	{
+		$generator = function() {
+			for ($i=0; $i<10; $i++) {
+				yield ($i+1);
+			}
+		};
+
+		$notAgenerator = function() {
```
jubianchi: `s/notAgenerator/notAGenerator`

tests/units/classes/asserters/generator.php
```diff
@@ -0,0 +1,129 @@
...
+			->then
+				->object($this->testedInstance->getGenerator())->isIdenticalTo($generator)
+				->object($this->testedInstance->getAnalyzer())->isIdenticalTo($analyzer)
+				->object($this->testedInstance->getLocale())->isIdenticalTo($locale)
+				->variable($this->testedInstance->getValue())->isNull()
+				->boolean($this->testedInstance->wasSet())->isFalse()
+		;
+	}
+
+	public function testUsage()
```
jubianchi: I think this kind of test should be in the `functionals` folder

agallou:
@jubianchi PR updated the generator asserter now as a returns assertion

Hywan:
One think is disturbing me. atoum syntax has always been: `$test->group($value)->asserter($params)`. Here, we only have `$test->asserter($params)`. The group here could be `yields` but this is not a type, while `group` mostly represents a type (or a category of types). So I would propose the following syntax:
```
$test->generator(…)->yields(…)->integer()->isEqualTo(…)
```
Note that `integer` has an implicit argument.

Thoughts?

agallou:
The syntax stays the same. The group is "generator" and the asserter is a variable asserter. The yields and returns keyworks a just here to simplify the tests on the asserters (it's kid of the the array asserter).

In your example : 
* what you the yields and integer calls would have as arguments ? 
* how would you distinguish an integeger call that is on the variable returned by yields and a "normal" asserter.

jubianchi:
@Hywan I think what you propose is the best we can provide, but for now, what @agallou did is good enough for a first version.

In his version, `$this->generator(/*...*/)->yields->isEqualTo(/*...*/)` unfolds to `$this->generator(/*...*/)->yields->variable->isEqualTo(/*...*/)`. The `variable` assertion is implied and we can keep this behavior, I think it's good.

In a second iteration, we can move to a more explicit API (and also keep the implied `variable`) .

mikaelrandy:
@jubianchi what you propose is to let people start to use a syntax, knowing that we will change this syntax in a near future.
I'm not sure it's a good idea. I prefer to wait for the "good" syntax

mikaelrandy:
But @agallou's work still an amazing work !

jubianchi:
@mikaelrandy we will be supporting both syntax: the implicit one and the explicit one.

agallou:
closing in favor of #664

PR #667, opened 2016-10-26
Add an asserter on phpArray to get the values

krtek4:
In the same vein that the [keys](http://docs.atoum.org/en/latest/asserters.html#keys) asserter this one allows you to easily get the values inside the array.


Hywan:
@jubianchi Can you restart the build on appveyor please?

jubianchi:
Done, everything is green!

PR #670, opened 2016-11-13
add sudo: false for travis

Grummfy:
close https://github.com/orgs/atoum/projects/1#card-777772


jubianchi:
`sudo: false` is already present in this file ;)

PR #672, opened 2016-11-21
3.0.0 remove closure

Grummfy:
For #641 one of the point is to remove the test pass to the closure.


Grummfy:
bad branch for the PR ;)

PR #678, opened 2016-12-20
Now, each mock instance may be unique, even if they are not identical.

mageekguy:
Hi folks,

That made a very loooooong time, and i'm very happy to contribute once again to `atoum`.  
Maybe you can see this PR as a gift for Xmas, with all my love ;), because you do an awesome job.  
However, now that I have say that, i can say that this PR may be seems weird at first look, but… i have very good reasons to do it.  
So, this is a test:

```
$this
	->given(
		$output = new mockOfOutput,

		$outputOfBlock1 = new mockOfOutput,
		$block1 = new mockOfBlock,
		$this->calling($block1)->blockArgumentsAre = function($output, $recipient) use ($outputOfBlock1) {
			$recipient->outputIs($outputOfBlock1);
		},

		$outputOfBlock2 = new mockOfOutput,
		$outputOfBlock2->id = uniqid(),
		$block2 = new mockOfBlock,
		$this->calling($block2)->blockArgumentsAre = function($output, $recipient) use ($outputOfBlock2) {
			$recipient->outputIs($outputOfBlock2);
		}
	)
	->if(
		$this->newTestedInstance($output)
	)
	->then
		->object($this->testedInstance->blockCollectionIs(new block\collection($block1, $block2)))
			->isEqualTo($this->newTestedInstance($outputOfBlock2))
		->mock($block1)
			->receive('blockArgumentsAre')
				->withArguments($output)
					->once
		->mock($block2)
			->receive('blockArgumentsAre')
				->withArguments($outputOfBlock1)
					->once
;
```

This test is perfectly valid at first look, but… there is a trap!  
In this context, `$output` is ALWAYS strictly equal to `$outputOfBlock2`, so to be clear, `$output == $outputOfBlock2`.  
So, `$this->newTestedInstance($output) == $this->newTestedInstance($outputOfBlock2)`, and assertion `->isEqualTo($this->newTestedInstance($outputOfBlock2))` is ALWAYS true.  
So, if i'm using `$output` instead of `$outputOfBlock2` if in the tested code, the test will be… GREEN, and not RED, whereas there is a bug in the tested code!  
And in this context, it's not possible to use `->withIdenticalArguments` to check that the return value was built with the right instance of `output` class.   
The only trick is to add `$outputOfBlock2->id = uniqid();` after `$outputOfBlock2` instantiation to be able to differentiate it from `$output`.  
And it's exactly that this PR do, but automatically.  
However, this change can introduce BC breaks in some tests, and we don't like BC breaks.  
So, this ID is added in mock instances only if `mock\generator::eachInstanceIsUnique()` was called somewhere during test setup.  
Happy code review!

### DIFF
--- a/classes/mock/generator.php
+++ b/classes/mock/generator.php
@@ -327,7 +327,7 @@ class generator
 
 			if ($this->eachInstanceIsUnique === true)
 			{
-				$mockedMethods .= "\t\t" . '$this->id = uniqid();' . PHP_EOL;
+				$mockedMethods .= self::generateUniqueId();
 			}
 
 			if (self::hasVariadic($constructor) === true)
@@ -786,7 +786,7 @@ class generator
 
 		if ($uniqueId === true)
 		{
-			$defaultConstructor .= "\t\t" . '$this->id = uniqid();' . PHP_EOL;
+			$defaultConstructor .= self::generateUniqueId();
 		}
 
 		$defaultConstructor .=
@@ -937,5 +937,10 @@ class generator
 			'while',
 			'xor',
 		);
+	}
+
+	private static function generateUniqueId()
+	{
+		return "\t\t" . '$this->{\'mock\' . uniqid()} = true;' . PHP_EOL;
 	}
 }
--- a/classes/mock/generator.php
+++ b/classes/mock/generator.php
@@ -20,6 +20,7 @@ class generator
 	protected $allowUndefinedMethodsUsage = true;
 	protected $allIsInterface = false;
 	protected $testedClass = '';
+	protected $eachInstanceIsUnique = false;
 
 	private $defaultNamespace = null;
 
@@ -140,6 +141,13 @@ class generator
 		return $this;
 	}
 
+	public function eachInstanceIsUnique()
+	{
+		$this->eachInstanceIsUnique = true;
+
+		return $this;
+	}
+
 	public function testedClassIs($testedClass)
 	{
 		$this->testedClass = strtolower($testedClass);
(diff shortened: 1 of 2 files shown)


classes/mock/generator.php
```diff
@@ -317,6 +325,11 @@ protected function generateClassMethodCode(\reflectionClass $class)
 			$mockedMethods .= PHP_EOL;
 			$mockedMethods .= "\t" . '{' . PHP_EOL;
 
+			if ($this->eachInstanceIsUnique === true)
+			{
+				$mockedMethods .= "\t\t" . '$this->id = uniqid();' . PHP_EOL;
```
jubianchi: what if my class has an actual `$id` property?

Hywan: A solution would be to generate a randomly named attribute:
```php
$this->{uniqid()} = true
```
Be careful that `uniqid()` must compute a valid PHP identifier. So maybe:
```php
$this->___atoum_{uniqid()} = true
```
Triple-underscore should be enought to avoid potential collision.

Hywan:
@mageekguy Maybe it would be good to target the `3.0…` branch instead of `master`. @jubianchi thoughts?

mageekguy:
From my point of view, its -1 for 3.0, because this PR does not introduce any BC break.  
The behaviour is different only if the user call `mock\generator::eachInstanceIsUnique()` (and somewhere, it's a bug…).

mageekguy:
Moreover, i have updated my PR according to @jubianchi's feedback (good catch, man, i'm always generate my mock as if mocked class is interface, so…).  
Instead of created a property `id` with a unique value, the mock generator add a unique property, aka `$this->{'mock' . uniqid()} = true`.

jubianchi:
+1 for targeting 2.9, as @mageekguy said there is no BC until the user enables the new option.

What we can discuss is making this a default behavior for 3.x.

PR #682, opened 2017-01-23
Fix bug in `mock\generator\allIsInterface()`.

mageekguy:
Before this commit, if `mock\generator\allIsInterface()` was called and a class is mocked, the parent class will be called in each method.
However, if `mock\generator\allIsInterface()` was used, it's because all mocks MUST be generated as if their base class is an interface (even if the base class is not an interface).
And… an interface has no method's implementation, so, calling the parent in this contexte is a non-sense.
So, this commit remove call to `parent::` if `mock\generator\allIsInterface()` was called.


Hywan:
It sounds good for me.

jubianchi:
Seems good but some tests have to be fixed ;)

Hywan:
:+1:

PR #697, opened 2017-02-19
Write an upgrade guide

jubianchi:


UPGRADE.md
```diff
@@ -0,0 +1,70 @@
+# atoum upgrade guide
+
+## From 2.x to 3.x
+
+### Runtime
+
+atoum `3.x` requires **PHP `>= 5.6.0`.
```
agallou: mettre `*PHP*` ?

UPGRADE.md
```diff
@@ -0,0 +1,70 @@
...
+    	$this->testedInstance->doSomethingAndThrow();
+    })
+;    
+```
+
+### Reports
+
+Some reports have been moved to a dedicated extension: [`atoum/reports-extension`](https://github.com/atoum/reports-extension).
+
+If you are using of of those reports, consider using the extension or simply remove them as they are not part of atoum anymore:
```
agallou: "one of those reports" ?

UPGRADE.md
Hywan: Or nothing. Why PHP should receive a specific semantics here?

jubianchi: Because I want to highlight the requirements.

PR #701, opened 2017-02-22
Allow the mock generator to use strict types

jubianchi:
Enabling this option will ensure mock receive and produce the right values.

Given the following class:

```php
<?php

namespace
{
    class foo
    {
        public function bar() : int
        {
            return rand(0, PHP_INT_MAX);
        }
    }
}
```

And the following test:

```php
<?php

namespace tests\units
{
    use atoum;

    class foo extends atoum
    {
        public function testBar(\foo $foo)
        {
            $this
                ->integer($foo->bar())
                    ->isGreaterThanOrEqualTo(0)
                    ->isLessThanOrEqualTo(PHP_INT_MAX)

                ->if($this->calling($foo)->bar = '42')
                ->then
                    ->integer($foo->bar())->isIdenticalTo(42)
            ;
        }
}
```

This test will pass without error.

But if we enable strict types:

```php
<?php

namespace tests\units
{
    use atoum;

    class foo extends atoum
    {
        public function testBar()
        {
            $this
                ->given(
                    $this->mockGenerator->useStrictTypes(),
                    $foo = new \mock\foo()
                )
                ->then
                    ->integer($foo->bar())
                        ->isGreaterThanOrEqualTo(0)
                        ->isLessThanOrEqualTo(PHP_INT_MAX)

                ->if($this->calling($foo)->bar = '42') // This is wrong
                ->then
                    ->integer($foo->bar())->isIdenticalTo(42)
            ;
        }
}
```

We will get a `TypeError: Return value of mock\foo::bar() must be of the type integer, string returned`

### DIFF
--- a/classes/mock/generator.php
+++ b/classes/mock/generator.php
@@ -19,6 +19,7 @@ class generator
     protected $allIsInterface = false;
     protected $testedClass = '';
     protected $eachInstanceIsUnique = false;
+    protected $useStrictTypes = false;
 
     private $defaultNamespace = null;
 
@@ -142,6 +143,13 @@ class generator
     public function eachInstanceIsUnique()
     {
         $this->eachInstanceIsUnique = true;
+
+        return $this;
+    }
+
+    public function useStrictTypes()
+    {
+        $this->useStrictTypes = true;
 
         return $this;
     }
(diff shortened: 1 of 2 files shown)


jubianchi:
I updated this PR to allow the mock generator to automatically generate valid `return`s (cc @mageekguy) :

```php
<?php

declare(strict_types=1);

namespace
{
    interface A
    {
        public function first() : self;
    }

    class B implements A
    {
        public function first() : A
        {
            return $this;
        }

        public function second() : self
        {
            return $this;
        }

        public function third() : int
        {
            return 42;
        }
    }

    class C extends B
    {
        public function second() : parent
        {
            return $this;
        }

        public function fourth() : ?int
        {
            return null;
        }

        public function fifth() : void
        {

        }
    }
}

namespace tests\units
{
    use atoum;

    class B extends atoum
    {
        public function testAll()
        {

            $this
                ->given($this->mockGenerator->useStrictTypes())
                ->assert('Mocking an interface')
                    ->if($a = new \mock\A())
                    ->then
                        ->object($a->first())->isIdenticalTo($a)

                ->assert('Mocking an actual class which implements an interface')
                    ->given($this->mockGenerator->allIsInterface())
                    ->if($b = new \mock\B())
                    ->then
                        ->object($b->first())->isIdenticalTo($b)
                        ->object($b->second())->isIdenticalTo($b)
                        ->exception(function() use ($b) { $b->third(); })
                            ->isInstanceOf(\TypeError::class)

                ->assert('Mocking an actual class which implements an interface and extends a class')
                    ->if($c = new \mock\C())
                    ->then
                        ->object($c->first())->isIdenticalTo($c)
                        ->object($c->second())->isIdenticalTo($c)
                        ->exception(function() use ($c) { $c->third(); })
                            ->isInstanceOf(\TypeError::class)
                        ->variable($c->fourth())->isNull
                        ->variable($c->fifth())->isNull
            ;
        }
    }
}
```

classes/mock/generator.php
```diff
@@ -559,32 +637,57 @@ protected function getNamespace($class)
         return '\\' . $this->getDefaultNamespace() . ($lastAntiSlash === false ? '' : '\\' . substr($class, 0, $lastAntiSlash));
     }
 
-    protected function getReturnType(\reflectionMethod $method)
+    protected function getReturnTypeCode(\reflectionMethod $method)
```
jubianchi: This method should not be renamed as it is used by the [`visibility-extension`}(https://github.com/atoum/visibility-extension/blob/master/classes/mock/generator.php#L47)

KuiKui:
Great ! Can we considered this PR as a bug fix ?

vdechenaux:
:+1: we can't use void type with atoum mock

```
vendor/atoum/atoum/classes/mock/generator.php(197) : eval()'d code on line 199:
A void function must not return a value
```

With these changes, it works

PR #704, opened 2017-03-13
Change success foreground color to make it more constrasted

guiled:
I always have a problem to easily read the success line because of the text color with present a low contrast with background (light-grey on green).
So I made a very minor change to make the text white.
I know this can be make with a configuration file (this is what I've done first locally), but as it concerns accessibilty of atoum, I think it may be more usefull as a change inside atoum itself.

### DIFF
--- a/classes/reports/realtime/cli.php
+++ b/classes/reports/realtime/cli.php
@@ -91,8 +91,8 @@ class cli extends realtime
 
         $runnerResultField = new runner\result\cli();
         $runnerResultField
-            ->setSuccessColorizer(new colorizer('0;37', '42'))
-            ->setFailureColorizer(new colorizer('0;37', '41'))
+            ->setSuccessColorizer(new colorizer('1;30', '42'))
+            ->setFailureColorizer(new colorizer('1;37', '41'))
         ;
 
         $this->addField($runnerResultField);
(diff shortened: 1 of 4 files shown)


classes/reports/realtime/cli.php
```diff
@@ -91,7 +91,7 @@ public function __construct()
 
         $runnerResultField = new runner\result\cli();
         $runnerResultField
-            ->setSuccessColorizer(new colorizer('0;37', '42'))
+            ->setSuccessColorizer(new colorizer('0;97', '42'))
```
jubianchi: This color code is out of the 8-bit color palette bounds. I'm afraid it won't work on some terminal emulators or environment (CI for example).

guiled: I agree... What about to choose 1;37 then ?
I'm not used to term colors...
Thank you

Hywan:
Because colors depend of the terminal, so is your terminal color palette the default one?

guiled:
Success with 0;37 (light gray, current default)


Success with 1;37 (bold gray, a more accessible way... 8bit color)


Success with 0;97 (real white, 16bit color)

jubianchi:
I pished a changes to the branch to use a more contrasted color :

* black on green for success
* white on red for failure

mikaelrandy:
Why do not use white on green, the same as for error output (white on red)

Results on my default iTerm color palette :
SuccessFailure

PR #705, opened 2017-03-24
Asserters/Stream::isWritten and Asserters/Stream::isRead can be used w/o brackets

guiled:
I made this PR in order to be consistent with isWritten and isRead like other asserters (ie Boolean::isTrue …)


jubianchi:
@guiled good for me. Could you also update the changelog please.

guiled:
changelog updated :-)

Issue #718, opened 2017-05-29
nullable type & mock : error


Grummfy:
Here is a small example with some useless test to prove that everything is ok
https://gist.github.com/Grummfy/bffb3bb5a1e8d6a8280f8879bf28cf4e

In short : mocking a class with nullable type (see http://php.net/manual/fr/migration71.new-features.php) as paramerter of method provocate errors

to fix it we need to detect the nullable type of paramter inside \mageekguy\atoum\mock\generator::getParametersSignature

PR #719, opened 2017-05-29
fix for atoum/atoum#718

Grummfy:
Here is the fix for #718 but honestly i'm not sure if can't do better with something around \mageekguy\atoum\mock\generator::getParameterType

But this code need to be compatible with previous version of php 7.1, so I think it fix everything.


classes/mock/generator.php
```diff
@@ -721,6 +721,7 @@ protected function getParametersSignature(\reflectionMethod $method, $forceMockC
                     $parameterCode .= ' = ' . var_export($parameter->getDefaultValue(), true);
                     break;
 
+                case $parameter->allowsNull():
```
Hywan: I guess this is not totally correct. For instance, given the function `f` signature:

```php
function f(?int $x, int $y);
```

The generated mock will be:

```php
function f(?int $x = null, int $y);
```

which is incorrect. We cannot have a required parameter *after* an optional parameter.

Grummfy: nope it will generate

`function f(int $x = null, int $y);`

the issue is that there is no way inside the reflection api to detect the nullable type. This was the only way to make it working (and it's a blocking bug)

Hywan: Well, the problem remains the same with `function f(int $x = null, int $y)`…

Grummfy: yes signature change, but ... we have no way to determine if it's a nullable type or not. So except if we change php core, we need a way to generate mocks... otherwise atoum will break all test with nullable type.

jubianchi: There is indeed no way to directly know if a parameter is nullable (`?`) or if it is `null` by default but by using some tricks you can get the correct information:

```php
<?php

class Foo
{
    public function bar(?int $a, int $b = null) {}
}

$method = new ReflectionMethod(Foo::class, 'bar');
$parameters = $method->getParameters();

function isNullable(ReflectionParameter $parameter) {
    return $parameter->allowsNull() && !$parameter->isDefaultValueAvailable();
}

function isDefaultNull(ReflectionParameter $parameter) {
    return $parameter->allowsNull() && $parameter->isDefaultValueAvailable();
}

var_dump(
    isNullable($parameters[0]), // bool(true)
    isDefaultNull($parameters[0]), // bool(false)
    isNullable($parameters[1]), // bool(false)
    isDefaultNull($parameters[1]) // bool(true)
);
```

Grummfy: well thanks for the code, I update the PR (and add a light improvement, we need to check that the default value != null ;)

by the way, don't you think it could be a nice issue to submit to php to have a isNullable() ?

jubianchi: @Grummfy another tiny improvement: https://3v4l.org/nUvuD

Grummfy:
Thanks a lot for the help, here is the fixed version. 

By the way, thanks to the atoum's day and the explanation of julien, I'm starting to find my way in atoum code ;)

classes/mock/generator.php
```diff
@@ -668,7 +668,7 @@ protected function getReturnType(\reflectionMethod $method)
         return $returnTypeCode;
     }
 
-    protected function isNullable(\reflectionType $type)
+    protected function isNullableReturnType(\reflectionType $type)
```
jubianchi: I would leave this as is because it could break extensions or any code extending this class.

classes/mock/generator.php
```diff
@@ -688,6 +688,18 @@ protected function isVoid(\reflectionMethod $method)
         return $this->hasReturnType($method) ? (string) $method->getReturnType() === 'void' : false;
     }
 
+    protected static function isNullableParameter(ReflectionParameter $parameter) {
+        return version_compare(PHP_VERSION, '7.1') >= 0 &&
+               $parameter->allowsNull() &&
+               (!$parameter->isDefaultValueAvailable() || ($parameter->isDefaultValueAvailable() && null !== $parameter->getDefaultValue()));
+    }
+
+	protected static function isDefaultParameterNull(ReflectionParameter $parameter) {
```
jubianchi: there seem to be an indentation error

classes/mock/generator.php
Grummfy: yep, but it' a bit confusing. So is there any space where we can note that kind of stuff for the v4?

Grummfy:
I have added 2 unit test for this one, copy-paste from others.
I also change a bit some other to fix code in relation with the new method used ;)

now it should be ready

Grummfy:
hello,
it could be helpfull to released (or at least to merge it with the master) this fix, it's a blocking one ;)

Hywan:
Thanks!

PR #726, opened 2017-07-11
fix(autoloader) Remove a cache warning

Hywan:
Remove a warning if the cache file is missing by using a simple `file_exists` instead of `@`.

Fix https://github.com/hoaproject/Kitab/issues/9.


jubianchi:
@Hywan seems good!

Could you add an entry tin the changelog for this bugfix ?

Once this is done, you can merge and release a new minor version of atoum.

Issue #696, opened 2017-02-14
Backport bugfixes from 3.0.0 to 2.x


jubianchi:
* [x] #688 - Report: Avoid reporting incorrect atoum path
* [x] #690 - Fix appveyor build
* [x] #691 - Fix how annotations are extracted. Only those actually starting with `@` are handled

PR #728, opened 2017-07-18
Fix 696

Grummfy:
fix for #696


Grummfy:
styleci failed is normal here ;)

Issue #725, opened 2017-07-10
function already exists


alex199568:
Trying to mock a function as was described in documentation. And all I get is the message that function already exists.

Here is my code

```

    public function testLogout()
    {
        $this
            ->assert('session unset and destroyed')
            ->given($this->newTestedInstance())
            ->if($this->function->session_unset) // check only unset for now
            ->then
            ->object($this->testedInstance->logout())
            ->isTestedInstance()
            ->function('session_unset')
            ->wasCalled()
            ->once();
    }

```

Here is the output

```
Failure (1 test, 3/3 methods, 0 void method, 0 skipped method, 0 uncompleted method, 0 failure, 1 error, 0 exception)!
> There is 1 error:
=> tests\units\Admin::testLogout():
==> Error E_USER_ERROR in C:\_\test\Admin.php on line 35, generated by file C:\_\test\Admin.php on line 35 in case 'session destroyed':
Function 'session_unset' already exists
 [Testing\Atoum]  Exit code 1  Time 5.214s

```

I don't think you'll be able to reproduce it, because this code is almost identical to the following snippet from documentation
```
<?php

$this
   ->assert('the file exist')
      ->given($this->newTestedInstance())
      ->if($this->function->file_exists = true)
      ->then
      ->object($this->testedInstance->loadConfigFile())
         ->isTestedInstance()
         ->function('file_exists')->wasCalled()->once()

   ->assert('le fichier does not exist')
      ->given($this->newTestedInstance())
      ->if($this->function->file_exists = false )
      ->then
      ->exception(function() { $this->testedInstance->loadConfigFile(); })
;
```

I just ran the above code from documentation and here is the result
```
Failure (1 test, 3/3 methods, 0 void method, 0 skipped method, 0 uncompleted method, 0 failure, 1 error, 0 exception)!
> There is 1 error:
=> tests\units\Admin::testLogout():
==> Error E_USER_ERROR in C:\_\test\Admin.php on line 45, generated by file C:\_\test\Admin.php on line 45 in case 'the file exist':
Function 'file_exists' already exists
 [Testing\Atoum]  Exit code 1  Time 1.012s

```

I have no idea what to do...

Hywan:
This is not possible to mock a function that has already been mock inside the same context (here, a test case). Is it your issue?

alex199568:
Thanks for the reply

I am mocking in one place only

I removed the line `->if($this->function->session_unset)` and simplified my test a little. Here is the updated test
```
$this
[...]

PR #736, opened 2017-08-31
feat: Display clear errors when mocking function fails

### LINKED ISSUE #735: Throw an exception when mocking a function in the root namespace
See #725.

When the user tries to mock a function for a class in the root namespace, atoum silently fails to do its work.

The failure is expected: it's impossible to mock a function in the root namespace, PHP will simply ignore it.

To make everything clear, atoum should throw an exception when the user attempts to do this so that everything is clear.

jubianchi:
This is to avoid question like the one in #725

Closes #735

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 # `dev-master`
 
+* [#736](https://github.com/atoum/atoum/pull/736) Display clear errors when mocking function fails ([@jubianchi])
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
 # `dev-master`
 
+* [#736](https://github.com/atoum/atoum/pull/736)  Display clear errors when mocking function fails" ([@jubianchi])
```
jubianchi:
```diff
- )  Display
+ ) Display
```

```diff
- fails"
+ fails
```

PR #741, opened 2017-09-07
add a way to assert on mocks arguments

agallou:
It would be usefull to run assertions on mock arguments,
like that we would be able to check for example a format like a datetime
and not the exact value.

Here is how it could look like :

```php
  $mock = new \mock\mageekguy\atoum\tests\functionals\asserters\objectAMocker();
  $mock->run('42', ['aaa']);
  $this
      ->mock($mock)
          ->call('run')
          ->onceAndCheckFirstCallArguments(function($arguments) {
              $this
                  ->string($arguments[0])
                      ->isEqualTo('42')
                      ->hasLength(2)
              ;
          })
;
```

There would be the same methods as usual but with a callback, and this callback has as first argument an array with the arguments passed to the asserted method. In this callback, the user will be able to do any assertion he needs.

This PR is not finished (for example unit tests are missing), it's a first draft to discuss how
 we could do that.

//cc @brunoroux for the original user need

### DIFF
--- a/classes/asserters/adapter/call.php
+++ b/classes/asserters/adapter/call.php
@@ -225,6 +225,42 @@ abstract class call extends atoum\asserter
         return $this;
     }
 
+    public function onceAndCheckFirstCallArguments(callable $test, $failMessage = null)
+    {
+        return $this->exactlyAndCheckFirstCallArguments($test, 1, $failMessage);
+    }
+
+    public function twiceAndCheckFirstCallArguments(callable $test, $failMessage = null)
+    {
+        return $this->exactlyAndCheckFirstCallArguments($test,2, $failMessage);
+    }
+
+    public function thriceAndCheckFirstCallArguments(callable $test, $failMessage = null)
+    {
+        return $this->exactlyAndCheckFirstCallArguments($test,3, $failMessage);
+    }
+
+    public function exactlyAndCheckFirstCallArguments(callable $test, $number, $failMessage = null)
+    {
+        $this->exactly($number, $failMessage);
+
+        /** @var \mageekguy\atoum\test\adapter\calls $calls */
+        $calls = $this->callIsSet()->adapter->getCalls($this->call, $this->identicalCall);
+
+        $callsArray = $calls->toArray();
+
+        /** @var \mageekguy\atoum\test\adapter\call $call */
+        $call = $callsArray[1];
+
+        if (!isset($callsArray[1])) {
+            $this->fail("Call not found");
+        }
+
+        $test(array_values($call->getArguments()));
+
+        return $this;
+    }
+
     public function never($failMessage = null)
     {
         return $this->exactly(0, $failMessage);
--- a/tests/functionals/classes/asserters/mock.php
+++ b/tests/functionals/classes/asserters/mock.php
@@ -1,0 +1,39 @@
+<?php
+
+namespace mageekguy\atoum\tests\functionals\asserters;
+
+use mageekguy\atoum
+;
+
+require_once __DIR__ . '/../../runner.php';
+
+class objectAMocker
+{
+    public function run($test) {
+
+    }
+}
+
+class mock extends atoum\tests\functionals\test\functional
+{
+    public function testUsage()
+    {
+        $mock = new \mock\mageekguy\atoum\tests\functionals\asserters\objectAMocker();
+        $mock->run('42', ['aaa']);
+        //$mock->run('42', ['aaa']);
+        $this
+            ->mock($mock)
+                ->call('run')
+                ->onceAndCheckFirstCallArguments(function($arguments) {
+                    $this
+                        ->string($arguments[0])
+                            ->isEqualTo('42')
+                            ->hasLength(2)
+                        ->array($arguments[1])
+                            ->hasSize(1)
+                    ;
+                })
+
+        ;
+    }
+}


Grummfy:
I'm not a fan of "onceAndCheckFirstCallArguments"
why not doing something like 
->received(function($arg1, $arg2)){$this->string($arg1);})
or 
->receivedAndCheck(function($arg1, $arg2)){$this->string($arg1);})

or something similar

jubianchi:
What about something like:

```php
<?php

$this
    ->mock($mock)
        ->call('foo')
            ->withArguments($this->arguments->string->isEqualTo('foo'), $this->arguments->any, $this->arguments->object)
        ->once
;
```

agallou:
@Grummfy yes the `onceAndCheckFirstCallArguments` is probably a little to verbose (I wanted to explain exacly what it does by with a long method name it could be less readable). So yes something like `onceAndCheck` will be easer to read and we can add some documentation to explain it's the argumernts on the first call that are tested.

@jubianchi I dont' understand how your example work, to call the asserters, the once method has to be called before (that's why have the callback). And how do the user selects which of the arguments is asserted ?

jubianchi:
@agallou while I find your proposal interesting, I can see some issues with it: first we would have to create several new methods :

* `onceAndCheckFirstCallArguments(callable $cb)`
* `twiceAndCheckFirstCallArguments(callable $cb)`
* `thriceAndCheckFirstCallArguments(callable $cb)`
* `atLeastAndCheckFirstCallArguments(int $number, callable $cb)`
* `exactlyAndCheckFirstCallArguments(int $number, callable $cb)`
* `neverAndCheckFirstCallArguments(callable $cb)`

Also, I don't know if you tested it with `before` and `after` assertions on mock calls:

```php
<?php

$this
    ->mock($a)
        ->call('foo')->onceAndCheckFirstCallArguments(function() {})
            ->before($this->mock($b)->call('bar')->onceAndCheckFirstCallArguments(function() {}))
;
` ``

Here is your first example translated with my proposal :


```php
<?php

$mock = new \mock\mageekguy\atoum\tests\functionals\asserters\objectAMocker();
$mock->run('42', ['aaa']);
$this
    ->mock($mock)
    ->call('run')->withArguments($this->arguments->string->isEqualTo(42)->hasLength(2))->once
;
```
_Here we are testing that the second argument is a string which equals to `42`._

To select the argument you want to assert on, you can use the `any` placeholder to skip some of them:

```php
<?php

$mock = new \mock\mageekguy\atoum\tests\functionals\asserters\objectAMocker();
$mock->run('42', ['aaa']);
$this
    ->mock($mock)
    ->call('run')->withArguments($this->arguments->any, $this->arguments->array->hasSize(1))->once
;
```

_Here we are testing that the second argument is an array with one element._

Grummfy:
Nice to see this synthax @jubianchi , for me is quiete clear and overall it keep the same idea as what exist.

But to improve it a bit, let's have these option in extra:
* a way to say that the first call to the method of the mock is with args foo & bar, and the second is baz & foobar.
* a way to say that there is at least this and this arguments

For information, here is the documentation of other tools:
mockery
* http://docs.mockery.io/en/latest/reference/expectations.html
* http://docs.mockery.io/en/latest/reference/argument_validation.html
phpunit
* https://phpunit.de/manual/current/en/test-doubles.html (not really intresting)
simpletest
* http://www.simpletest.org/fr/mock_objects_documentation.html
peridot, behat, phpspec => prophecy or mockery
kahlan I don't find a way to define something in relation with the arguments of the mock

btry:
Hi

I feel not familiar enough with atoum as I very recently migrated from phpunit to atoum (and so learn it at the same time). Then I don't assert my proposal is the definite good way. Anyway here are my thought.

I suggest the role of the method ```onceAndCheckFirstCallArguments``` (or any other name mentioned here) be splitted in two distinct methods.

As atoum uses a fluent interface I would prefer something like this :
```php
$mock = new \mock\mageekguy\atoum\tests\functionals\asserters\objectAMocker();
  $mock->run('42', ['aaa']);
  $this
      ->mock($mock)
          ->call('run')
          ->once()
          ->checkFirstCallArguments(function($arguments) {
              $this
                  ->string($arguments[0])
                      ->isEqualTo('42')
                      ->hasLength(2)
              ;
          })
;
```

Applied to @jubianchi 's suggestion, this would split  each method he submitted into 2 others; each having only one responsibility. I guess this would also help to maintain the feature by reducing the size of the new methods.

jubianchi:
@Grummfy 

> a way to say that the first call to the method of the mock is with args foo & bar, and the second is baz & foobar.

```php
<?php

$this
    ->mock($mock)
    // Assertions on first call
    ->call('run')
        ->withArguments($this->arguments->string->isEqualTo('foo'), $this->arguments->string->isEqualTo('bar'))
            ->before(
                $this->mock($mock)
                    // Assertions on second call
                    ->call('run')->withArguments($this->arguments->string->isEqualTo('baz'), $this->arguments->string->isEqualTo('foobar'))->once
            )
        ->once
;
```

> a way to say that there is at least this and this arguments

* http://docs.atoum.org/en/latest/asserters.html#withatleastarguments
* http://docs.atoum.org/en/latest/asserters.html#withatleastidenticalarguments

jubianchi:
This PR will be closed in favor of #744.

See [here](https://github.com/atoum/atoum/pull/744#issuecomment-333533228) for the final proposal.

PR #744, opened 2017-09-29
feat: Assert on mock arguments

jubianchi:
In #741 @agallou proposed something to test mock arguments in a more convenient way.

This reminded me about #254 which I could not get merged (too much conflicts to resolve at some point).

I propose an alternative implementation [here](https://github.com/atoum/atoum/pull/741#issuecomment-332015867). This PR provides the implementation for this proposal.

```php
<?php

namespace {
    class bar {
        public function foo($a) {}
    }
}

namespace tests\units {
    use atoum;

    class bar extends atoum {
        public function testLogout()
        {
            $this
                ->given($mock = new \mock\bar())
                ->when($mock->foo([$a = uniqid()]))
                ->then
                    ->mock($mock)->call('foo')->withArguments($this->args->array->string[0]->isEqualTo($a))->once
            ;
        }
    }
}
```

### DIFF
--- a/classes/asserters/adapter.php
+++ b/classes/asserters/adapter.php
@@ -54,6 +54,11 @@ class adapter extends call
         return $this->withAtLeastArguments([]);
     }
 
+    public function verify(callable $verify)
+    {
+        return $this->setVerify($verify);
+    }
+
     protected function adapterIsSet()
     {
         try {
--- a/classes/asserters/adapter/call.php
+++ b/classes/asserters/adapter/call.php
@@ -312,6 +312,7 @@ abstract class call extends atoum\asserter
             ->call
                 ->setFunction($function)
                 ->unsetArguments()
+                ->unsetverify()
         ;
 
         $this->beforeCalls = [];
@@ -328,6 +329,7 @@ abstract class call extends atoum\asserter
             ->setTrace()
             ->call
                 ->setArguments($arguments)
+                ->unsetverify()
         ;
 
         $this->identicalCall = false;
@@ -343,6 +345,37 @@ abstract class call extends atoum\asserter
             ->setTrace()
             ->call
                 ->unsetArguments()
+        ;
+
+        $this->identicalCall = false;
+
+        return $this;
+    }
+
+    protected function setVerify(callable $verify)
+    {
+        $this
+            ->adapterIsSet()
+            ->callIsSet()
+            ->setTrace()
+            ->call
+                ->setVerify($verify)
+                ->unsetArguments()
+        ;
+
+        $this->identicalCall = false;
+
+        return $this;
+    }
+
+    protected function unsetVerify()
+    {
+        $this
+            ->adapterIsSet()
+            ->callIsSet()
+            ->setTrace()
+            ->call
+                ->unsetVerify()
         ;
 
         $this->identicalCall = false;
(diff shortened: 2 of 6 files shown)


Grummfy:
perhaps we can combine the feature from @agallou and this one. I like this synthax, but the use of an anonymous function is also seducing, because it permit to use more complex test. Can we use both, like 
->mock($mock)->call('foo')->withArguments($this->args->...
->mock($mock)->call('foo')->withArguments(function(){}) or event an other assertion.

jubianchi:
@Grummfy @agallou what about this:

```php
<?php

namespace {
    class bar {
        public function foo($a) {}
    }
}

namespace tests\units {
    use atoum;

    class bar extends atoum {
        public function testLogout()
        {
            $this
                ->given($mock = new \mock\bar())
                ->when($mock->foo([$a = uniqid()]))
                ->then
                    ->mock($mock)
                        ->call('foo')->withArguments($this->args->verify(function($arguments) use ($a) {
                            $this->array($arguments)
                                ->hasSize(1)
                                ->string[0]->isEqualTo($a)
                            ;
                        }))->once
            ;
        }
    }
}

```

Grummfy:
```
 ->call('foo')->withArguments($this->args->integer, $this->args->verify(function($arguments) use ($a) {
                            $this->array($arguments)
                                ->hasSize(1)
                                ->string[0]->isEqualTo($a)
                            ;
                        }))->once
```
is possible?
if yes, I think this cover all case. Let's wait @agallou

agallou:
```php
                        ->call('foo')->withArguments($this->args->verify(function($arguments) use ($a) {
                            $this->array($arguments)
                                ->hasSize(1)
                                ->string[0]->isEqualTo($a)
                            ;
                        }))->once
```

@jubianchi yes, i find this version easier to read than the one on the PR description. the callbacks indicates clearly that the assertion will be delayed, and that will give the user the possibility to do pretty much everything he wants has assertions.

jubianchi:
@agallou @Grummfy here is my final implementation. I reworked the whole thing to keep only the `verify` part.

Here is an example:

```php
<?php

namespace {
    class bar {
        public function foo($a) {}
    }
}

namespace tests\units {
    use atoum;

    class bar extends atoum {
        public function testLogout()
        {
            $this
                ->given($mock = new \mock\bar())
                ->when($mock->foo([$a = uniqid()]))
                ->then
                    ->mock($mock)
                        ->call('foo')->verify(function($arguments) use ($a) {
                            $this->array($arguments)
                                ->hasSize(1)
                                ->array[0]
                                    ->isEqualTo([$a])
                            ;
                        })->once
        ;
        }
    }
}
```

Grummfy:
in term of usage : perfect for me.

agallou:
@jubianchi i've made some tests and my first feedback is that it works great : thanks.
I will do some more tests during the week (to test some edge cases and the error messages).

For info here is how we were getting the arguments to assert on them at @Decitre-Interactive :
```php
$callArray = $this->mock($mock)->call('foo')->once()->getAdapter()
    ->getCalls(new \mageekguy\atoum\test\adapter\call('foo'))
    ->toArray()
;
$callArray = array_values($callArray);
$arguments = $callArray[0]->getArguments();
//then to the assertions on this array
```
The new way to do it will be much easier and will feel less like a hack.

Grummfy:
seems good, so we can merge it, no?

Grummfy:
@jubianchi is it finished for you?
@Hywan or @mikaelrandy ready for release?

PR #747, opened 2017-09-29
chore: Add a upper bound to the PHP constraint

jubianchi:


Grummfy:
why?

jubianchi:
@Grummfy because we don't know in advance what PHP will eventually break in future versions.

We had several examples where we had to add upper bounds to constraint after some "breaking" changes were introduced :

* `void` classes had to be renamed;
* `string` classes had to be renamed;
* `integer` classes had to be renamed;
* ...

Grummfy:
oki

PR #752, opened 2017-10-02
feat: Add an os annotation to only run tests on specific OS

jubianchi:
Here is an example :

```php
namespace tests\units
{
    /**
     * @os !darwin
     */
    class stdClass extends \mageekguy\atoum\test
    {
        /**
         * @os linux
         */
        public function testOnlyOnLinux()
        {

        }

        /**
         * @os !linux
         */
        public function testEverywhereButLinuxAndDarwin()
        {

        }
    }
}
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 # `dev-master`
 
+* [#752](https://github.com/atoum/atoum/pull/752) Add an os annotation to only run tests on specific OS ([@jubianchi])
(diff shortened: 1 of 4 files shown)


Grummfy:
seems perfect in terms of usage ;)

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 # 3.2.0 - 2017-09-07
 
+* [#752](https://github.com/atoum/atoum/pull/752) Add an os annotation to only run tests on specific OS ([@jubianchi])
```
Grummfy: bad place in changelog ;)

PR #757, opened 2017-10-09
fix(runner) Use error reporting level to exit

Hywan:
The error reporting level can exclude some error levels. Take this into account before exiting the runner.


vonglasow:
:+1: a little bit late ;)

PR #769, opened 2017-10-10
feat(script) Align options to the left, and increase contrast

### LINKED ISSUE #760: CLI: Help text options are hard to read
I know this is my opinion, but I find the center-alignment of the help text options (when running `atoum --help`) difficult to read. It's also a pattern I've never seen in other CLI tools which are typically left-aligned and wrap in columns.

The CLI help text also has no maximum line-length and when it wraps it's hard to read. It would be better to wrap within columns.

Finally, the default colors for the options have poor contrast on a black background. The common pattern I've seen in other CLI tools is to not colorize the help text.

For example here's my help text:

screen shot 2017-10-09 at 7 00 45 pm

Compared to, for example, Jest:

screen shot 2017-10-09 at 7 01 40 pm

Hywan:
Fix #760.

Result:
screen shot 2017-10-10 at 10 54 22

Thoughts @sirbrillig?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 # `dev-master`
 
+* [#769](https://github.com/atoum/atoum/pull/769) CLI: Align options to the left, and increase contrast ([@hywan])
 * [#770](https://github.com/atoum/atoum/pull/770) Fix path to the runner for the help ([@hywan])
 * [#768](https://github.com/atoum/atoum/pull/768) Fix typos in the CLI help ([@hywan])
 * [#767](https://github.com/atoum/atoum/pull/767) Fix typos in the exception messages ([@hywan])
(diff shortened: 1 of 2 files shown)


vonglasow:
@Hywan seems ok, have you try on small screen also to see if it looks well also ? If yes good for me.

Hywan:
@vonglasow On a small screen, it looks waayyyyy better than previously :-).

Hywan:
Oops, too late…

jubianchi:
@Hywan please, open a new PR to fix the tests and the color.

Hywan:
@jubianchi Sure!

Hywan:
@jubianchi https://github.com/atoum/atoum/pull/772

Hywan:
@mageekguy Agree, and a low contrast impact everyone. Now we have a much better contrast, that's one step foward.

PR #771, opened 2017-10-10
feat(asserter) Normalize the asserter name

### LINKED ISSUE #765: Assertion failures show full path for assertion without assertion name
Assertion failures use full paths for the assertion, which seems like it is maybe not necessary, especially for common assertions (e.g.:  `isEqualTo()` produces `mageekguy\atoum\asserters\phpString() failed: strings are not equal`). 

Furthermore, it does not use the name of the assertion (`isEqualTo`) anywhere in the error.

Hywan:
Fix #765.

  * Remove the `mageekguy\atoum\asserters` prefix if it exists,
  * Remove the `php` prefix is it exists.

### DIFF
--- a/classes/asserter/exception.php
+++ b/classes/asserter/exception.php
@@ -31,15 +31,14 @@ class exception extends \runtimeException
                 }
             }
 
-            $asserterName = strtolower(get_class($asserter));
-
-            if ('mageekguy\atoum\asserters\\' === substr($asserterName, 0, 26)) {
-                $asserterName = substr($asserterName, 26);
-            }
-
-            if (0 !== preg_match('/^php./', $asserterName)) {
-                $asserterName = substr($asserterName, 3);
-            }
+            $asserterName = preg_replace(
+                [
+                    '/^' . preg_quote('mageekguy\atoum\asserters\\') . '/',
+                    '/^php(?=.)/'
+                ],
+                '',
+                strtolower(get_class($asserter))
+            );
 
             $code = $test->getScore()->addFail($file, $class, $method, $line, $asserterName . ($function ? '::' . $function : '') . '()', $message);
         }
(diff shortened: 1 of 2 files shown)


classes/asserter/exception.php
```diff
@@ -31,7 +31,17 @@ public function __construct(atoum\asserter $asserter, $message)
...
             }
 
-            $code = $test->getScore()->addFail($file, $class, $method, $line, get_class($asserter) . ($function ? '::' . $function : '') . '()', $message);
+            $asserterName = strtolower(get_class($asserter));
+
+            if ('mageekguy\atoum\asserters\\' === substr($asserterName, 0, 26)) {
+                $asserterName = substr($asserterName, 26);
+            }
+
+            if (0 !== preg_match('/^php./', $asserterName)) {
```
jubianchi: why not simply calling `preg_replace` here instead of a ìf`+`substr` ?

classes/asserter/exception.php
```diff
@@ -31,7 +31,17 @@ public function __construct(atoum\asserter $asserter, $message)
                 }
             }
 
-            $code = $test->getScore()->addFail($file, $class, $method, $line, get_class($asserter) . ($function ? '::' . $function : '') . '()', $message);
+            $asserterName = strtolower(get_class($asserter));
+
+            if ('mageekguy\atoum\asserters\\' === substr($asserterName, 0, 26)) {
```
jubianchi: why not simply calling preg_replace here instead of a `if`+`substr`s ?

Hywan: Agree. Done.

classes/asserter/exception.php
Hywan: Agree. Done.

PR #773, opened 2017-10-11
fix(script) Directory is the current working directory

Hywan:
Explicitely use the current working directory the script directory.

This behavior was already present before, but a bug fix (see #770) has revealed this other bug.

The script directory must not rely on the script name (which is defined in the `atoum` binary by the `atoum\scripts\runner` constant). The script name points to the binary name most of the time (it is the case right now), but the script directory is the user current working directory, they are not related.

### DIFF
--- a/classes/script.php
+++ b/classes/script.php
@@ -46,11 +46,7 @@ abstract class script
 
     public function getDirectory()
     {
-        $directory = $this->adapter->dirname($this->getName());
-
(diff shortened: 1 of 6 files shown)


Hywan:
There is an issue where `phar::running()` returns an empty string. It seems to not be related to this patch though…

Hywan:
@jubianchi Tests are greens, ready for a review. Only StyleCI is red for some reasons.

PR #774, opened 2017-10-11
fix(php) Fix a syntax error…

Hywan:


jubianchi:
@Hywan what is the syntax error here? I don't see anything wrong in the code before your patch.

PR #777, opened 2017-10-30
cli output changed for displaying errors and exceptions clearly

macintoshplus:
I have add some colors and new line for clarify the errors and exceptions in console output.
Sample of result :
change default report

This simple separator and color is helpful for resolve error when run tests.


Grummfy:
seems far clearer than before. thanks
I let @jubianchi or @hywan validate it

Hywan:
Hello,

Thanks for the PR! We are constantly fixing the color scheme. I think it's better to come with a solution proposed by @mageekguy in https://github.com/atoum/atoum/pull/769#commitcomment-24879725: A customizable color scheme profile. Also, I think the default report is not very easy to read, and should be revamped.

It is possible to develop custom report with extensions. It might be a direction to take. Thoughts @jubianchi?

Grummfy:
for the color, I agree, but for the format, I think this is very helpfull, we all know that this need to be improved. At least it's a very start here

macintoshplus:
What do I have to change to merge?

Hywan:
@macintoshplus It's not clear yet if we will accept your PR as is, or if we will come with a different solution to customize the color scheme :-).

macintoshplus:
Ok define a color scheme it's a good idea, but howto get the color from the scheme ?

Grummfy:
For me, this is ready. It's not perfect for the whole solution, but it's better than what we have now. So I would like that we accept it. It's a nice contribution, and it's a required one.

Grummfy:
@Hywan @jubianchi @mikaelrandy for me we should accept it, as a first step.

Grummfy:
hey @macintoshplus  I know it's a while since it's here, but because I want to merge it, I'm wondering if you want to fix it or if you are ok that I fix it and merge it. Since we didn't move on anything in the rest, this deserve to be merged, it's a nice first step.

Grummfy:
merged through https://github.com/atoum/atoum/pull/833

PR #785, opened 2017-12-23
like for exception, now output can use assertion directly

Grummfy:
Instead of writing this:
```php
$this
    ->output(function(atoum\test $test) { 
        $test->testedInstance->doSomethingAndEcho();
    })
;
```
we can now write this:

```php
$this
    ->output(function() { 
        $this->testedInstance->doSomethingAndEcho();
    })
;
```

### DIFF
--- a/classes/asserters/output.php
+++ b/classes/asserters/output.php
@@ -19,7 +19,11 @@ class output extends phpString
     {
         if ($value instanceof \closure) {
             ob_start();
-            $value($this->getTest());
+            if (version_compare(PHP_VERSION, '7.0.0') >= 0) {
+                $value($this->getTest());
+            } else {
+                $value();
+            }
             $value = ob_get_clean();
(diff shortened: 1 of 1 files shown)


Hywan:
Don't forget to drop a line in the `CHANGELOG.md` :-).

mageekguy:
It’s already possible to not inject a test instance in the anonymous function. This PR is, sadly, useless.

mikaelrandy:
@mageekguy can you give an exemple ?
That interest me !

jubianchi:
as @mageekguy this PR does not bring anything new. In fact, we can already write 

```php
$this
    ->output(function() { 
        $this->testedInstance->doSomethingAndEcho();
    })
;
```

This is because PHP will automatically bind `$this` in anonymous functions. Keeping the `$test` argument will ensure backward compatibility.

We could remove the `$test` argument, as we did everywhere else (AFAIR), to be consistent.

Issue #790, opened 2018-02-26
allIsInterface is not reseted after mock


Grummfy:
an example:
```php
<?php

namespace
{

	class Foo
	{
		public function getFoo(): int
		{
			return 42;
		}
	}

	class FooBar extends Foo
	{
		public function getFoo(): int
		{
			return parent::getFoo() + 1;
		}
	}

	class FooBaz extends Foo
	{
		public function getFoo(): int
		{
			return parent::getFoo() - 1;
		}
	}

	class Yolo
	{
		public function olyo(Foo $foo)
		{
			return $foo->getFoo();
		}
	}
}

namespace test\unit
{
	class Yolo extends \atoum\test
	{
		public function testMock()
		{
			$mock1 = new \mock\FooBaz();
			$mock2 = new \mock\FooBar();
			$this->newTestedInstance();
			$this->integer($this->testedInstance->olyo($mock1))->isEqualTo(41);
			$this->integer($this->testedInstance->olyo($mock2))->isEqualTo(43);
		}

		public function testMockInterface()
		{
			$mock1 = new \mock\FooBaz();
			$this->mockGenerator->allIsInterface();
			$mock2 = new \mock\FooBar();
			$this->calling($mock2)->getFoo = 43;
			$this->newTestedInstance();
			$this->integer($this->testedInstance->olyo($mock1))->isEqualTo(41);
			$this->integer($this->testedInstance->olyo($mock2))->isEqualTo(43);
		}

		public function testMockInterfaceFail()
		{
			$this->mockGenerator->allIsInterface();
			$mock2 = new \mock\FooBar();
			$this->calling($mock2)->getFoo = 43;

			$mock1 = new \mock\FooBaz();
			$this->newTestedInstance();
			$this->integer($this->testedInstance->olyo($mock1))->isEqualTo(41);
			$this->integer($this->testedInstance->olyo($mock2))->isEqualTo(43);
		}
	}
}
```

error generated

```
$ vendor/bin/atoum -f test.php 
> atoum path: .../vendor/atoum/atoum/bin/atoum
> atoum version: 3.2.0
> PHP path: /usr/bin/php7.1
> PHP version:
=> PHP 7.1.11-0ubuntu0.17.10.1 (cli) (built: Nov  1 2017 16:30:52) ( NTS )
=> Copyright (c) 1997-2017 The PHP Group
=> Zend Engine v3.1.0, Copyright (c) 1998-2017 Zend Technologies
[...]

PR #791, opened 2018-02-26
fix: allIsInterface not reseted #790

Grummfy:
fix for #790


Grummfy:
I'm wondering if it's the same behaviour for eachInstanceIsUnique

mageekguy:
Hi guys!
This behaviour is not a bug.
`allIsInterface` is a global configuration directive for the mock generator, because it is the philosophical opposite of its default behavior.
Default behaviour is not true OOP, because it allow the developper to build code based on implementation.
The `allIsInterface` behaviour "force" the developper to build code based upon true abstraction (aka interface), so it enforce the user to do true OOP.
So, in my opinion, allow the user to mix this two philosophie is not the way to go, and it is the reason why `allIsInterface` is a global configuration directive, instead of a local directive which may be reset after mock instantiation.
Morevover, this PR is a BC break, because tests suite which was configured to use `allIsInterface` will be silently switch to "implementation based" mode.
In conclusion, it's a "no go" for me for this PR.
Have a good day!

PR #792, opened 2018-03-02
Take care to reset relative URL to prevent issues; fixes #783

### LINKED ISSUE #783: HTML report sometimes add extra parent path
Sometimes (I cannot really tell when it happens), some generated coverage HTML files have a wrong path. This affetcs the top link and the path to the CSS file as well (well, at least).

On a correct generated file, I'll have `src="screen.css"` but `src="../../screen.css"` on wrong ones.

trasher:


classes/report/fields/runner/coverage/html.php
```diff
@@ -116,6 +116,8 @@ public function __toString()
 
                     if (substr_count($className, '\\') >= 1) {
                         $classTemplate->relativeRootUrl = rtrim(str_repeat('../', substr_count($className, '\\')), DIRECTORY_SEPARATOR) . DIRECTORY_SEPARATOR;
+                    } else {
+                        $classTemplate->relativeRootUrl = '';
```
Hywan: I think `'./'` might be a little bit more consistent than `''`. Thoughts?

Hywan:
Thanks for the patch!

classes/report/fields/runner/coverage/html.php
trasher: It seems `relativeRootUrl` is empty by default; that's why I've used `''`. I've just tested with `./`; it works as well. Lemme know what you prefer, I'll change.

Hywan: I prefer `./` :-).

trasher: Changed

Issue #794, opened 2018-03-09
atoum exits when the bootstrap fails


Hywan:
Hello fellow atoum hackers,

I faced an issue today. On a code we don't control entirely, the PHP version has been updated. PHP now emits errors for undefined constants. That's expect, everything is normal. However, we can't run the tests anymore because atoum exits before running the tests, as defined here:

https://github.com/atoum/atoum/blob/b25d916d21ac20b958e0cd55a52af41625ad5763/classes/scripts/runner.php#L691-L699

Maybe it's not a good idea to exit the process directly. Logging a message is enough I think. It's very likely that the errors will be reported by the tests too. In our situation, we prefer to see the tests running, and use reports to see the errors, than not being able to run the tests.

Thoughts?

cc @jubianchi @Grummfy

Grummfy:
as I know it exit from the test and that's all, no? Otherwie we have a behaviour not coherent with the other exit, no?

Hywan:
No, it exits if a bootsrap error happens. For an error raised during a test executed, it's reported as is, and the runner continues to execute tests.

PR #795, opened 2018-03-09
Don't exit when an error is reported during the bootstrap

Hywan:
See https://github.com/atoum/atoum/issues/794 for the discussion.


Hywan:
Travis is failing because it can't install atoum.

PR #801, opened 2018-03-17
Use spaces indentation in composer.json

petk:
Hello, this patch is a CS fix to move from tabs to spaces. In case of `composer.json` files mostly 4 spaces are used in PHP OS projects...

Thanks for considering checking this out or merging it.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,65 +1,65 @@
 {
-	"name": "atoum/atoum",
-	"type": "library",
-	"description": "Simple modern and intuitive unit testing framework for PHP 5.3+",
-	"keywords": ["TDD","atoum","test","unit testing"],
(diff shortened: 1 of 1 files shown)


Grummfy:
I don't see the intrest of it

petk:
It's a consistency patch and developers usually want that in their code base. I haven't seen tabs used in the `composer.json` files before. There are used in json files anywhere important?

Also since the project doesn't have the .editorconfig configuration, manually editing composer.json file will cause spaces on most of today's editors I think.

But that's up to you to decide...

* https://blog.sourced.tech/post/tab_vs_spaces/

petk:
Besides, this is opened for 25 days now, I think the tabs are the way to go here. Sorry to bother you with my time.

PR #830, opened 2019-06-21
Add 7.4 to travis

remicollet:


.travis.yml
```diff
@@ -5,6 +5,7 @@ php:
   - 7.1
   - 7.2
   - 7.3
+  - 7.4snapshot
```
jubianchi: As PHP 7.4 is not release yet, I would make it an allowed failure for now.

What do you think ?

jubianchi:
Thanks @remicollet !

.travis.yml
trasher: Allowed failure or not, could you please merge this one?
This will be useful in order to work on 7.4 compatibility... Thanks!

PR #835, opened 2019-10-04
Fix  deprecated. unparenthesized (php 7.4)

trasher:
Fix following issue with php 7.4:
```
Fix Unparenthesized `a ? b : c ?: d` is deprecated. Use either `(a ? b : c) ?: d` or `a ? b : (c ?: d)`
``


trasher:
follows #827

PR #850, opened 2020-10-21
Fix calling instance method as static method

idetinkin:


jubianchi (approved):

Grummfy:
thanks for your contribution
Merci pour ta contribution

PR #883, opened 2022-08-09
Mock generator correctly handles the `static` return type

shavounet:
Since Symfony 6, the [`static` return type](https://www.php.net/manual/fr/language.types.declarations.php#language.types.declarations.static) is sometime used (see [Request](https://github.com/symfony/http-foundation/blob/5f0f9ef2e714ea5b2038a73fbf26bc75c04a4a6a/Request.php#L436) for example).
Atoum do not handle it properly : it generates a `: \static` method signature which fails...
```
==> Error FATAL ERROR in ... on unknown line, generated by file /src/vendor/atoum/atoum/classes/mock/generator.php(219) : eval()'d code on line 66:
'\static' is an invalid class name
```

This PR should fix it by using the same behavior as [`self`](https://github.com/atoum/atoum/pull/633).

I also applied latest CS fixer rules (as required by this package), which changed a few files... This is included in another commit, if you prefer to create 2 independent PR. It seems that it doesn't fix the CI though :disappointed: 

Note that the `never` return type is also missing, but not included in this PR.

### DIFF
--- a/classes/asserters/adapter/call.php
+++ b/classes/asserters/adapter/call.php
@@ -31,18 +31,18 @@ abstract class call extends atoum\asserter
             return $this->exactly($property);
         } else {
             switch (strtolower($property)) {
-            case 'once':
-            case 'twice':
-            case 'thrice':
-            case 'never':
-            case 'atleastonce':
-            case 'wascalled':
-            case 'wasnotcalled':
-                return $this->{$property}();
+                case 'once':
+                case 'twice':
+                case 'thrice':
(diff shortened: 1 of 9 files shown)


cedric-anne (approved):

fferriere (approved):

Oliboy50 (approved):

Grummfy (approved):

Grummfy:
Thanks for the work.
bonne journée

PR #897, opened 2024-02-06
Remove deprecated dynamic property usage

fberthereau:
I found this issue while trying to run my test suite on PHP 8.2

- Deprecated creation of dynamic property `$runnerTestsCoverageField` in `atoum\atoum\reports\realtime\cli\travis`

To address this, I replaced the dynamic property with a local variable since I couldn't find any usage of it.


Grummfy (changes requested):
if possible avoid changing code style

Grummfy (approved):

